PR #2, opened 2023-12-12
initial commit

zane-neo:
### Description
This is the initial commit to skills repo that contains only a PPLTool.

### Issues Resolved
This is the initial commit to skills repo that contains only a PPLTool.


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -0,0 +1,318 @@
...
+                ActionRequest request = new MLPredictionTaskRequest(
+                    modelId,
+                    MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(inputDataSet).build()
+                );
+                client.execute(MLPredictionTaskAction.INSTANCE, request, ActionListener.<MLTaskResponse>wrap(mlTaskResponse -> {
+                    ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlTaskResponse.getOutput();
+                    ModelTensors modelTensors = modelTensorOutput.getMlModelOutputs().get(0);
+                    ModelTensor modelTensor = modelTensors.getMlModelTensors().get(0);
+                    Map<String, String> dataAsMap = (Map<String, String>) modelTensor.getDataAsMap();
+                    String ppl = dataAsMap.get("output");
```
joshuali925, line 117: I tested the PPL tool with question `what are the errors?` on sample logs index, and the response says 
```
{
  "error": {
    "root_cause": [
      {
        "type": "exception",
        "reason": "execute ppl: source=\\`opensearch_dashboards_sample_data_logs\\` | where QUERY_STRING(['response'], '4* OR 5*')\n\nQuestion: What are the errors in the last 5 minutes? index is \\`opensearch_dashboards_sample_data_logs\\`, get error: Failed to parse query due to offending symbol [Question:] at: ' source=\\`opensearch_dashboards_sample_data_logs\\` | where QUERY_STRING(['response'], '4* OR 5*')\n\nQuestion:' <--- HERE...
```

It seems everything in the LLM output is considered as PPL?

In the langchain implementation, we ask LLM to use `<ppl>` tags, and throw errors if there is no `<ppl>` tag in the response. This will prevent issues in my example, where the question also became part of LLM output and was used as PPL query. It will also guard against invalid questions where PPL query cannot be generated.

Will there be plans to implement equivalent features as the langchain version?

greaa-aws: Want to add a +1 here.

Parsing LLM output is a key step, certainly for PPL-gen and likely for other tools as well. For PPL-gen, we know that the parsing steps implemented in the langchain version are essential to achieve its current level of performance. These steps include  the PPL tags ([ref](https://github.com/opensearch-project/dashboards-assistant/blob/e6359634ff6f532afa983ae0785bc0cd53aee425/server/olly/chains/ppl_generator.ts#L247)), ISNOTNULL catch ([ref](https://github.com/opensearch-project/dashboards-assistant/blob/e6359634ff6f532afa983ae0785bc0cd53aee425/server/olly/chains/ppl_generator.ts#L252)), backtick removal ([ref](https://github.com/opensearch-project/dashboards-assistant/blob/e6359634ff6f532afa983ae0785bc0cd53aee425/server/olly/tools/tool_sets/ppl.ts#L150)), and SPAN catch ([ref](https://github.com/opensearch-project/dashboards-assistant/blob/e6359634ff6f532afa983ae0785bc0cd53aee425/server/olly/tools/tool_sets/ppl.ts#L151)). We may see significantly degraded output quality without these steps implemented.

PR #39, opened 2023-12-19
Add basic search detectors tool; pull plugin deps in gradle run

ohltyler:
### Description
This PR adds 2 things:

- Adds basic implementation for part of the search detectors tool. Remaining functionality will be added when the profile API is onboarded in the AD client
- Adds dependencies on AD and job scheduler
- Improves `./gradlew run` to run an integ test cluster with ML commons, SQL, AD, and job scheduler plugins for faster iterative testing

Confirmed all build / test / run cmds work successfully on 2.11. The only difference for `main` is bumping to 3.0. Given the dependency list I'm not enforcing 3.0 builds to pass.

Additionally, I will manually open a backport to 2.x with the correct version
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ configurations {
     all {
         resolutionStrategy {
             force "org.mockito:mockito-core:5.8.0"
-            force "com.google.guava:guava:32.1.2-jre" // CVE for 31.1
+            force "com.google.guava:guava:32.1.3-jre" // CVE for 31.1
             force("org.eclipse.platform:org.eclipse.core.runtime:3.30.0") // CVE for < 3.29.0, forces JDK17 for spotless
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -12,14 +12,6 @@ buildscript {
         opensearch_version = System.getProperty("opensearch.version", "2.11.0-SNAPSHOT")
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
         buildVersionQualifier = System.getProperty("build.version_qualifier", "")
-        version_tokens = opensearch_version.tokenize('-')
-        opensearch_build = version_tokens[0] + '.0'
-        if (buildVersionQualifier) {
-            opensearch_build += "-${buildVersionQualifier}"
-        }
-        if (isSnapshot) {
-            opensearch_build += "-SNAPSHOT"
-        }
     }
 
     repositories {
@@ -113,7 +105,7 @@ dependencies {
     implementation("com.google.guava:guava:32.1.3-jre")
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
     
-    zipArchive "org.opensearch.plugin:opensearch-anomaly-detection:${opensearch_build}"
+    zipArchive "org.opensearch.plugin:opensearch-anomaly-detection:${version}"
     implementation fileTree(dir: adJarDirectory, include: ["opensearch-anomaly-detection-${version}.jar"])
 
     compileOnly "org.opensearch:common-utils:${version}"
@@ -127,7 +119,7 @@ dependencies {
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.10.1'
 
     // ZipArchive dependencies used for integration tests
-    zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${version}"
 }
 
 task extractSqlJar(type: Copy) {
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ import java.util.concurrent.Callable
 buildscript {
     ext {
         opensearch_group = "org.opensearch"
-        opensearch_version = System.getProperty("opensearch.version", "3.0.0-SNAPSHOT")
+        opensearch_version = System.getProperty("opensearch.version", "2.11.0-SNAPSHOT")
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
         buildVersionQualifier = System.getProperty("build.version_qualifier", "")
         version_tokens = opensearch_version.tokenize('-')
@@ -77,6 +77,7 @@ apply plugin: 'opensearch.testclusters'
 apply plugin: 'opensearch.pluginzip'
 
 def sqlJarDirectory = "$buildDir/dependencies/opensearch-sql-plugin"
+def adJarDirectory = "$buildDir/dependencies/opensearch-anomaly-detection"
 
 configurations {
     zipArchive
@@ -96,6 +97,11 @@ task addJarsToClasspath(type: Copy) {
         include "protocol-${version}.jar"
     }
     into("$buildDir/classes")
+
+    from(fileTree(dir: adJarDirectory)) {
+        include "opensearch-anomaly-detection-${version}.jar"
+    }
+    into("$buildDir/classes")
 }
 
 dependencies {
@@ -106,6 +112,10 @@ dependencies {
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
     implementation("com.google.guava:guava:32.1.3-jre")
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
+    
+    zipArchive "org.opensearch.plugin:opensearch-anomaly-detection:${opensearch_build}"
+    implementation fileTree(dir: adJarDirectory, include: ["opensearch-anomaly-detection-${version}.jar"])
+
     compileOnly "org.opensearch:common-utils:${version}"
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
     testImplementation "org.mockito:mockito-core:5.8.0"
@@ -126,7 +136,14 @@ task extractSqlJar(type: Copy) {
     into sqlJarDirectory
 }
 
+task extractAdJar(type: Copy) {
+    mustRunAfter()
+    from(zipTree(configurations.zipArchive.find { it.name.startsWith("opensearch-anomaly-detection")}))
+    into adJarDirectory
+}
+
 tasks.addJarsToClasspath.dependsOn(extractSqlJar)
+tasks.addJarsToClasspath.dependsOn(extractAdJar)
 project.tasks.delombok.dependsOn(addJarsToClasspath)
 tasks.publishNebulaPublicationToMavenLocal.dependsOn ':generatePomFileForPluginZipPublication'
 tasks.validateNebulaPom.dependsOn ':generatePomFileForPluginZipPublication'
@@ -161,6 +178,7 @@ spotless {
 
 compileJava {
     dependsOn extractSqlJar
+    dependsOn extractAdJar
     dependsOn delombok
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
 }
@@ -168,6 +186,8 @@ compileJava {
 compileTestJava {
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
 }
+
+forbiddenApisTest.ignoreFailures = true
 
 
 opensearchplugin {
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ apply plugin: 'opensearch.pluginzip'
 
 def sqlJarDirectory = "$buildDir/dependencies/opensearch-sql-plugin"
 def jsJarDirectory = "$buildDir/dependencies/opensearch-job-scheduler"
-def adJarDirectory = "$buildDir/dependencies/opensearch-anomaly-detection"
+def adJarDirectory = "$buildDir/dependencies/opensearch-time-series-analytics"
 
 configurations {
     zipArchive
@@ -97,7 +97,7 @@ task addJarsToClasspath(type: Copy) {
     into("$buildDir/classes")
 
     from(fileTree(dir: adJarDirectory)) {
-        include "opensearch-anomaly-detection-${version}.jar"
+        include "opensearch-time-series-analytics-${version}.jar"
     }
     into("$buildDir/classes")
 }
@@ -112,7 +112,7 @@ dependencies {
     // Plugin dependencies
     compileOnly group: 'org.opensearch', name:'opensearch-ml-client', version: "${version}"
     implementation fileTree(dir: jsJarDirectory, include: ["opensearch-job-scheduler-${version}.jar"])
-    implementation fileTree(dir: adJarDirectory, include: ["opensearch-anomaly-detection-${version}.jar"])
+    implementation fileTree(dir: adJarDirectory, include: ["opensearch-time-series-analytics-${version}.jar"])
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
     compileOnly "org.opensearch:common-utils:${version}"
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -77,13 +69,15 @@ apply plugin: 'opensearch.testclusters'
...
+def jsJarDirectory = "$buildDir/dependencies/opensearch-job-scheduler"
+def adJarDirectory = "$buildDir/dependencies/opensearch-anomaly-detection"
 
 configurations {
     zipArchive
     all {
         resolutionStrategy {
             force "org.mockito:mockito-core:5.8.0"
-            force "com.google.guava:guava:32.1.3-jre" // CVE for 31.1
+            force "com.google.guava:guava:32.1.2-jre" // CVE for 31.1
```
dbwiddis, line 80: Can we add a comment why we are not using the latest version?  Mend will generate an auto-bump for this (if it hasn't already) and would like to include a reason when closing it to prevent future bumps...

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Override
+    public boolean validate(Map<String, String> parameters) {
+        return true;
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
```
dbwiddis, line 154: This duplicates the Lombok `@Getter` annotation for the `type` field which I don't see actually used
```
    @Getter
    private String type;
```
One of these two pieces of code is wrong, probably the other one.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,202 @@
...
+    private static final String DEFAULT_DESCRIPTION = "Use this tool to search anomaly detectors.";
+
+    @Setter
+    @Getter
+    private String name = TYPE;
+    @Getter
+    @Setter
+    private String description = DEFAULT_DESCRIPTION;
+    @Getter
+    private String type;
```
dbwiddis, line 45: This dupicates the `getType()` method below, and seems to be otherwise unused, so I suspect this should be deleted.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,202 @@
...
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        final String detectorName = parameters.getOrDefault("detectorName", null);
+        final String detectorNamePattern = parameters.getOrDefault("detectorNamePattern", null);
+        final String indices = parameters.getOrDefault("indices", null);
+        final Boolean highCardinality = parameters.containsKey("highCardinality")
+            ? Boolean.parseBoolean(parameters.get("highCardinality"))
+            : null;
+        final Long lastUpdateTime = parameters.containsKey("lastUpdateTime") ? Long.parseLong(parameters.get("lastUpdateTime")) : null;
+        final String sortOrderStr = parameters.getOrDefault("sortOrder", "asc");
+        final SortOrder sortOrder = sortOrderStr == "asc" ? SortOrder.ASC : SortOrder.DESC;
```
dbwiddis, line 86: We may want to be case-insensitive for this particular parameter value.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,202 @@
...
+
+        private static Factory INSTANCE;
+
+        /** 
+         * Create or return the singleton factory instance
+         */
+        public static Factory getInstance() {
+            if (INSTANCE != null) {
+                return INSTANCE;
+            }
```
dbwiddis, line 172: I know you just copied this from pretty much all the other tools (including one I wrote that may have started the trend) but I think this whole double-checked-locking bit is overkill for this factory method.  But may as well leave it here for consistency!

src/test/java/org/opensearch/agent/tools/SearchAnomalyDetectorsToolTests.java
```diff
@@ -0,0 +1,152 @@
...
+            new SearchResponseSections(new SearchHits(hits, totalHits, 0), new Aggregations(new ArrayList<>()), null, false, null, null, 0),
+            null,
+            0,
+            0,
+            0,
+            0,
+            null,
+            null
+        );
+        String expectedResponseStr = String.format("AnomalyDetectors=[]TotalAnomalyDetectors=%d", hits.length);
```
dbwiddis, line 84: (Multiple occurrences) Best practice to use a Locale with `.format()`, even if it's the default locale.  

Less important as it's in a test, and the ones in this test are probably ok but could in theory cause a flaky test if numbers are rendered differently so I just do it out of habit.

dbwiddis (commented):
A few nits.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
dbwiddis: Also don't use `==` for Strings.  Use `.equals()` or per my suggestion `.equalsIgnoreCase()`.

zane-neo:
Build is failing, seems like missing dependency.

ohltyler:
> Build is failing, seems like missing dependency.

See comment in description:
Confirmed all build / test / run cmds work successfully on 2.11. The only difference for main is bumping to 3.0. Given the dependency list I'm not enforcing 3.0 builds to pass.

It is unsustainable to maintain passing builds on main since most plugin builds fail due to constant breaking changes on `main`/3.0. We can ensure passing builds on 2.x.

build.gradle
ohltyler: Thanks for the catch. Reverted this as it's a leftover change from my inter-plugin dependency issues.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Thanks for the catch - removing

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Removing

ohltyler: Appears to be a common pattern with the existing tools I've noticed

ohltyler: Looks like a leftover from the type refactoring previously done.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Oops! Still breaking out of my typescript habits :)

src/test/java/org/opensearch/agent/tools/SearchAnomalyDetectorsToolTests.java
ohltyler: Thanks - I've updated. Timestamp CI failures are indeed painful

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
dbwiddis: Yeah I had the same thing happen with CatIndices when I moved it to main. Those yellow warnings in my IDE were helpful.

build.gradle
```diff
@@ -12,14 +12,6 @@ buildscript {
...
         isSnapshot = "true" == System.getProperty("build.snapshot", "true")
         buildVersionQualifier = System.getProperty("build.version_qualifier", "")
-        version_tokens = opensearch_version.tokenize('-')
-        opensearch_build = version_tokens[0] + '.0'
-        if (buildVersionQualifier) {
-            opensearch_build += "-${buildVersionQualifier}"
-        }
-        if (isSnapshot) {
-            opensearch_build += "-SNAPSHOT"
-        }
```
dbwiddis, line 22: I'm not sure why these need to be removed. Commenting to be sure it's intentional.

build.gradle
```diff
@@ -96,28 +90,51 @@ task addJarsToClasspath(type: Copy) {
...
 
 dependencies {
-    compileOnly group: 'org.opensearch', name:'opensearch-ml-client', version: "${version}"
+    // 3P dependencies
     compileOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     compileOnly "org.apache.logging.log4j:log4j-slf4j-impl:2.22.0"
     compileOnly group: 'org.json', name: 'json', version: '20231013'
-    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
-    implementation("com.google.guava:guava:32.1.3-jre")
+    implementation("com.google.guava:guava:32.1.2-jre")
```
dbwiddis, line 110: This is still left over from your exploration. There's an open renovate PR for a version bump to 33.0.0, though.

build.gradle
```diff
@@ -96,28 +90,51 @@ task addJarsToClasspath(type: Copy) {
...
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
+
+    // Test dependencies
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
-    testImplementation "org.mockito:mockito-core:5.8.0"
+    testImplementation group: 'junit', name: 'junit', version: '4.13.2'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '5.8.0'
+    testImplementation group: 'org.mockito', name: 'mockito-inline', version: '5.2.0'
+    testImplementation("net.bytebuddy:byte-buddy:1.14.7")   
+    testImplementation("net.bytebuddy:byte-buddy-agent:1.14.7")
```
dbwiddis, line 131: Nit: while all the different formats are fine, I'd prefer to see all the dependencies listed in the same format (quoted in a single string without parentheses.   Won't hold up a PR merge for it, just makes things easier to find a dependency scanning the list.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,201 @@
...
+
+package org.opensearch.agent.tools;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.ad.client.AnomalyDetectionNodeClient;
```
dbwiddis, line 14: Test is failing saying it can't find this.

dbwiddis (commented):
LGTM with comments

build.gradle
ohltyler: Duplicate of lines 46-55

build.gradle
ohltyler: Oh right - it's included in 2 places in build.gradle and didn't revert this one, will just change back to 32.1.3 to limit the changes.

build.gradle
ohltyler: Fair enough, will keep in mind. This was already inconsistent so I think it can be fully cleaned up when the repo itself & its deps become more stable.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Right. It is because of the difficulty in having passing builds in 3.0 at all, especially with all of the added plugin dependencies. See comment above https://github.com/opensearch-project/skills/pull/39#issuecomment-1864815319

ohltyler: To prevent friction and keep things moving along, prefer to enforce on 2.x only. There is lots of changes happening in dependent repos that are inevitably going to keep breaking things here given the dependency model.

dbwiddis (approved):
LGTM with previous comments noted.  However, not sure we should merge something that doesn't compile and breaks `main`.  That will kill all testing of future PRs.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
dbwiddis: We need to set up end to end testing and my understanding is that will be on `main`.  Merging this would prevent any other CI checks from completing on any future PRs here.

ohltyler:
> LGTM with previous comments noted. However, not sure we should merge something that doesn't compile and breaks `main`. That will kill all testing of future PRs.

I've root caused this failure - the plugin name change has changed the underlying jar path in the zip, hence not adding it to the classpath. I also got AD to build and upload to sonatype so the snapshot is now successfully pulled in CI.

dbwiddis (approved):

dbwiddis:
> the plugin name change 

Ah, yes!  Should have remembered that.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,201 @@
...
+    // future to cover a sufficient amount of potential questions the agent will need to handle.
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        final String detectorName = parameters.getOrDefault("detectorName", null);
+        final String detectorNamePattern = parameters.getOrDefault("detectorNamePattern", null);
+        final String indices = parameters.getOrDefault("indices", null);
+        final Boolean highCardinality = parameters.containsKey("highCardinality")
+            ? Boolean.parseBoolean(parameters.get("highCardinality"))
+            : null;
+        final Long lastUpdateTime = parameters.containsKey("lastUpdateTime") ? Long.parseLong(parameters.get("lastUpdateTime")) : null;
```
zane-neo, line 83: `Long.parseLong` might throw NumberFormatException if the input parameter is null or invalid number format, suggestion is to use `StringUtils.isNumeric()` to check if the input is a number.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -0,0 +1,201 @@
...
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
+            .query(boolQueryBuilder)
+            .size(size)
+            .from(startIndex)
+            .sort(sortString, sortOrder);
+
+        SearchRequest searchDetectorRequest = new SearchRequest().source(searchSourceBuilder);
+
+        if (running != null || disabled != null || failed != null) {
+            // TODO: add a listener to trigger when the first response is received, to trigger the profile API call
```
zane-neo, line 122: Is this missed to address or will be addressed in future?

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
zane-neo: Agree, currently all getInstance() method are invoked in plugin initialization which is single thread. And here is a bug even the assumption is multiple threads will access to this method, which is: the INSTANCE is missing `volatile` decorator and another thread could be using a partial constructed object and causing exception.

dbwiddis: Which is why I should have reflexively objected to double-checked-locking at the start.  This probably needs a repo-wide fix to just wrap everything in a synchronized method and be done with it. It's not hot-path.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Will be addressed in the future. I'll create a follow up issue alongside some other improvements soon.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Thanks! I will update.

zane-neo (approved):

PR #40, opened 2023-12-20
feature: Add AbstractRetriverTool, VectorDBTool, NeuralSparseTools

zhichao-aws:
### Description
Move the AbstractRetriverTool, VectorDBTool, NeuralSparseTools from ml-commons agent_framework_dev branch to skills.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,7 +121,7 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
-    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${opensearch_build}"
 }
 
 task extractSqlJar(type: Copy) {
@@ -147,7 +147,6 @@ test {
         showStandardStreams true
     }
     include '**/*Tests.class'
-    systemProperty 'tests.security.manager', 'false'
 }
 
 spotless {
--- a/build.gradle
+++ b/build.gradle
@@ -104,9 +104,9 @@ dependencies {
     compileOnly "org.apache.logging.log4j:log4j-slf4j-impl:2.22.0"
     compileOnly group: 'org.json', name: 'json', version: '20231013'
     compileOnly "org.opensearch:common-utils:${version}"
+    compileOnly("com.google.guava:guava:32.1.2-jre")
+    compileOnly group: 'org.apache.commons', name: 'commons-lang3', version: '3.10'
 
-    implementation("com.google.guava:guava:32.1.3-jre")
-    implementation group: 'org.apache.commons', name: 'commons-lang3', version: "${versions.commonslang}"
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
 
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
--- a/build.gradle
+++ b/build.gradle
@@ -200,8 +200,6 @@ compileJava {
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
 }
 
-forbiddenApisTest.ignoreFailures = true
-
 compileTestJava {
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
 }
--- a/build.gradle
+++ b/build.gradle
@@ -107,8 +107,8 @@ dependencies {
     compileOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     compileOnly "org.apache.logging.log4j:log4j-slf4j-impl:2.22.0"
     compileOnly group: 'org.json', name: 'json', version: '20231013'
-    implementation("com.google.guava:guava:32.1.3-jre")
-    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.13.0'
+    compileOnly("com.google.guava:guava:32.1.3-jre")
+    compileOnly group: 'org.apache.commons', name: 'commons-lang3', version: '3.10.0'
 
     // Plugin dependencies
     compileOnly group: 'org.opensearch', name:'opensearch-ml-client', version: "${version}"
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ configurations {
     zipArchive
     all {
         resolutionStrategy {
-            force "org.mockito:mockito-core:5.8.0"
+            force "org.mockito:mockito-core:${versions.mockito}"
             force "com.google.guava:guava:32.1.3-jre" // CVE for 31.1
             force("org.eclipse.platform:org.eclipse.core.runtime:3.30.0") // CVE for < 3.29.0, forces JDK17 for spotless
         }
@@ -103,12 +103,15 @@ dependencies {
     compileOnly group: 'com.google.code.gson', name: 'gson', version: '2.10.1'
     compileOnly "org.apache.logging.log4j:log4j-slf4j-impl:2.22.0"
     compileOnly group: 'org.json', name: 'json', version: '20231013'
-    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
-    implementation("com.google.guava:guava:32.1.3-jre")
-    implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
     compileOnly "org.opensearch:common-utils:${version}"
+
+    implementation("com.google.guava:guava:32.1.3-jre")
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: "${versions.commonslang}"
+    implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
+
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
-    testImplementation "org.mockito:mockito-core:5.8.0"
+    testImplementation "net.bytebuddy:byte-buddy-agent:${versions.bytebuddy}"
+    testImplementation "org.mockito:mockito-core:${versions.mockito}"
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.10.1'
     testImplementation 'org.mockito:mockito-junit-jupiter:5.8.0'
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+        this.client = client;
+        this.xContentRegistry = xContentRegistry;
+        this.index = index;
+        this.sourceFields = sourceFields;
+        this.docSize = docSize == null ? DEFAULT_DOC_SIZE : docSize;
+    }
+
+    protected abstract String getQueryBody(String queryText);
+
+    private <T> SearchRequest buildSearchRequest(Map<String, String> parameters, ActionListener<T> listener) throws IOException {
```
mingshl, line 69: does buildSearchRequest depend on listener?

zhichao-aws: Thanks for point out, will remove it

zhichao-aws:
UT are finished, the integration test are still blocked by an exception during cluster bootstrap:
```
org.opensearch.bootstrap.StartupException: java.lang.IllegalStateException: Plugin [skills] cannot extend non-extensible plugin [opensearch-ml]
```
This issue depends on the ml-commons plugin. I think we can review and merge this PR first to unblock other developers. After ml-commons fix the issue, we can add integration test in follow up PRs.

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+@Log4j2
+@Getter
+@Setter
+public abstract class AbstractRetrieverTool implements Tool {
+    public static final String DEFAULT_DESCRIPTION = "Use this tool to search data in OpenSearch index.";
+    public static final String INPUT_FIELD = "input";
+    public static final String INDEX_FIELD = "index";
+    public static final String SOURCE_FIELD = "source_field";
+    public static final String DOC_SIZE_FIELD = "doc_size";
+    public static final int DEFAULT_DOC_SIZE = 2;
```
zane-neo, line 43: Is this to control the matching docs size? I remember it's 10 by default in search API. Are we giving 2 randomly or there's a reason behind this?

zhichao-aws: This number is inherited from origin VectorDBTool. I guess the reason maybe we want to limit the context size for LLM. I'm open for changing this number

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+        return searchRequest;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        SearchRequest searchRequest;
+        try {
+            searchRequest = buildSearchRequest(parameters);
+        } catch (Exception e) {
+            listener.onFailure(e);
```
zane-neo, line 91: Add error log here?

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+
+        ActionListener actionListener = ActionListener.<SearchResponse>wrap(r -> {
+            SearchHit[] hits = r.getHits().getHits();
+
+            if (hits != null && hits.length > 0) {
+                StringBuilder contextBuilder = new StringBuilder();
+                for (int i = 0; i < hits.length; i++) {
+                    SearchHit hit = hits[i];
+                    Map<String, Object> docContent = new HashMap<>();
+                    docContent.put("_index", hit.getIndex());
```
zane-neo, line 103: Is there any specific reason to extract these fields instead of deserialize the whole searchResponse as result?

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+                    docContent.put("_id", hit.getId());
+                    docContent.put("_score", hit.getScore());
+                    docContent.put("_source", hit.getSourceAsMap());
+                    contextBuilder.append(gson.toJson(docContent)).append("\n");
+                }
+                listener.onResponse((T) contextBuilder.toString());
+            } else {
+                listener.onResponse((T) "");
```
zane-neo, line 111: Do we need to return some prompt to user/thought so that user can understand the index search doesn't find any match?

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,143 @@
...
+        client.search(searchRequest, actionListener);
+    }
+
+    @Override
+    public boolean validate(Map<String, String> parameters) {
+        if (parameters == null || parameters.size() == 0) {
+            return false;
+        }
+        String question = parameters.get("input");
+        return !StringUtils.isBlank(question);
```
zane-neo, line 126:
```suggestion
        return parameters != null && parameters.size() >= 0 && !StringUtils.isBlank(parameters.get("input"));
```

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
zhichao-aws: Sure

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
zhichao-aws: Good point.

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
zhichao-aws: https://github.com/opensearch-project/OpenSearch/blob/2c8ee1947b55a1cc5bc1a114b82e3a3b8a99851e/server/src/main/java/org/opensearch/search/SearchHit.java#L616
If we just deserialize the object, the unnecessary fields may increase the size of context. E.g.  nestedIdentity, version, seqNo, primaryTerm etc.

mingshl (approved):
@zhichao-aws overall it looks good to me even though the codecov is failing.

I wrote some tests to increase code coverage for AbstactRetrieverTool, we can merge this PR first and I will help contribute the codecov in a new PR.

https://github.com/mingshl/ml-commons/commit/93c729a122fd4b6624bd219c26d95dd167d8a983

build.gradle
```diff
@@ -118,6 +121,7 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${opensearch_build}"
+    zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
```
dhrubo-os, line 124: Shouldn't all the plugins have the same version with `opensearch_build`?

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -0,0 +1,140 @@
...
+                    Map<String, Object> docContent = new HashMap<>();
+                    docContent.put("_index", hit.getIndex());
+                    docContent.put("_id", hit.getId());
+                    docContent.put("_score", hit.getScore());
+                    docContent.put("_source", hit.getSourceAsMap());
+                    contextBuilder.append(gson.toJson(docContent)).append("\n");
+                }
+                listener.onResponse((T) contextBuilder.toString());
+            } else {
+                listener.onResponse((T) "Can not get any match from search result.");
```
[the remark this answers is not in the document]
zhichao-aws: Based on my mental model the retrieval tools are used for RAG, so their response will be parsed by LLM instead of human. We don't have a tool-retry logic now, the "refine your search terms" may confuse LLM

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
zhichao-aws: The origin VectorDBTool was introduced from the agent framework POC commit. https://github.com/opensearch-project/ml-commons/commit/1b85cffb07977a3aedb61a41637cb1b24f6cd016#diff-b353ce1eda5b942809ea500dadcfda5769504edd8fa17fc174d498937097c69eR67 Hi @ylwu-amzn, do you know why the default doc size is 2? Do we set this for context length?

By the way I don't think this is a blocker issue for this PR. This parameter is configurable, and we can alter its value after our e2e test.

build.gradle
```diff
@@ -82,7 +82,7 @@ configurations {
     zipArchive
     all {
         resolutionStrategy {
-            force "org.mockito:mockito-core:5.8.0"
+            force "org.mockito:mockito-core:${versions.mockito}"
```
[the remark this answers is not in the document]
zhichao-aws: It's 5.5.0 in this case
https://github.com/opensearch-project/OpenSearch/blob/f92f846a1f9b30a055dde846fd12d987a511723a/buildSrc/version.properties#L58

build.gradle
zhichao-aws: makes sense

build.gradle
```diff
@@ -165,6 +170,8 @@ compileJava {
     options.compilerArgs.addAll(["-processor", 'lombok.launch.AnnotationProcessorHider$AnnotationProcessor'])
 }
 
+forbiddenApisTest.ignoreFailures = true
```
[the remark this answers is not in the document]
zhichao-aws: It's inherited from ml-commons build script. I try to build without this line and get errors like
```
Forbidden annotation use: org.junit.Test [defaultMessage Just name your test method testFooBar]
  in org.opensearch.agent.tools.NeuralSparseSearchToolTests (NeuralSparseSearchToolTests.java, annotation on method declaration of 'testCreateTool()')
```
I tried to do some research but can not get any useful information.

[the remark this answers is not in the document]
zhichao-aws: And if I removes the @Test annotation, it just can not find the test case

build.gradle
```diff
@@ -137,12 +141,13 @@ testingConventions.enabled = false
...
 
 test {
-    useJUnitPlatform()
     testLogging {
         exceptionFormat "full"
         events "skipped", "passed", "failed" // "started"
         showStandardStreams true
     }
+    include '**/*Tests.class'
+    systemProperty 'tests.security.manager', 'false'
```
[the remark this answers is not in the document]
zhichao-aws: It turns off the security manager to make it easy for writing test case. It is used in many build script of Opensearch plugins. However we don't have such test cases currently. So I just removed this line and we can add it back when we need it.

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
zane-neo: The result looks like in jsonl format, is this as expected?

zhichao-aws: Yes, this result will be treated as a norm string and be sent to LLM.

zane-neo (approved):

PR #52, opened 2023-12-22
Add a search AD results tool

ohltyler:
### Description
This PR adds a search AD results tool. It exposes several different parameters to search and sort the desired results. Note these are subject to change as more integration testing is done with agents configured with the tool.

Testing:
- manually tested the generated queries against a test cluster to ensure they work as expected against different scenarios and different types of anomaly results
- adds limited UT consistent with existing coverage of tools for now. Again this may be improved when the inputs/outputs/parameters become more standardized and we can have more robust coverage and detailed test cases. We can see how the performance is when we benchmark the tools in integration tests with an equipped agent configured with an LLM
 
### Issues Resolved
https://github.com/opensearch-project/ml-commons/issues/1548
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
@@ -70,7 +70,6 @@ public class SearchAnomalyResultsTool implements Tool {
         };
     }
 
-    // TODO: update description
     // Response is currently in a simple string format including the list of anomaly results (only detector ID, grade, confidence),
     // and toal # of results. The output will likely need to be updated, standardized, and include more fields in the
     // future to cover a sufficient amount of potential questions the agent will need to handle.
--- a/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
@@ -139,8 +139,8 @@ public class SearchAnomalyResultsTool implements Tool {
             sb.append("AnomalyResults=[");
             for (SearchHit hit : hits) {
                 sb.append("{");
-                sb.append("id=").append(hit.getId()).append(",");
-                sb.append("grade=").append(hit.getSourceAsMap().get("anomaly_grade"));
+                sb.append("detectorId=").append(hit.getSourceAsMap().get("detector_id")).append(",");
+                sb.append("grade=").append(hit.getSourceAsMap().get("anomaly_grade")).append(",");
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
```diff
@@ -0,0 +1,210 @@
...
+
+    // Response is currently in a simple string format including the list of anomaly results (only detector ID, grade, confidence),
+    // and toal # of results. The output will likely need to be updated, standardized, and include more fields in the
+    // future to cover a sufficient amount of potential questions the agent will need to handle.
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        final String detectorId = parameters.getOrDefault("detectorId", null);
+        final Boolean realTime = parameters.containsKey("realTime") ? Boolean.parseBoolean(parameters.get("realTime")) : null;
+        final Double anomalyGradeThreshold = parameters.containsKey("anomalyGradeThreshold")
+            ? Double.parseDouble(parameters.get("anomalyGradeThreshold"))
```
dhrubo-os, line 81: can we add sufficient test for this method?

zane-neo, line 81: Do we need to check if the parameters.get("anomalyGradeThreshold") is a number before parsing to Double?

ohltyler: `isNumeric` requires each character to be a digit, but users may want to pass `-1` for example. Prefer to leave as-is, as `-1` may be a common value (within LLM prompt potentially) to indicate they want to fetch all anomaly results, regardless if they are actual anomalies with grades > 0.

ohltyler: Added

dhrubo-os: can't we add more info in the test like `anomalyGradeThreshold`, `dataStartTime`, `dataEndTime` etc?

ohltyler: Added for other valid cases with the other params. But there won't be any other invalid cases since the other params are tested with `isNumeric`, which if it fails, will default to `null` values and not throw exceptions.

dhrubo-os (approved):

zane-neo (approved):

PR #57, opened 2023-12-25
Add UT and parser for ppl tool

xinyual:
### Description
Add UT and output parser for PPL tool according to the suggestion from @joshuali925 
 
### Issues Resolved
Add UT and output parser for PPL tool
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -17,6 +17,8 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.StringJoiner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.json.JSONObject;
 import org.opensearch.action.ActionRequest;
@@ -116,7 +118,7 @@ public class PPLTool implements Tool {
                     ModelTensors modelTensors = modelTensorOutput.getMlModelOutputs().get(0);
                     ModelTensor modelTensor = modelTensors.getMlModelTensors().get(0);
                     Map<String, String> dataAsMap = (Map<String, String>) modelTensor.getDataAsMap();
-                    String ppl = dataAsMap.get("response");
+                    String ppl = parseOutput(dataAsMap.get("response"), indexName);
                     JSONObject jsonContent = new JSONObject(ImmutableMap.of("query", ppl));
                     PPLQueryRequest pplQueryRequest = new PPLQueryRequest(ppl, jsonContent, null, "jdbc");
                     TransportPPLQueryRequest transportPPLQueryRequest = new TransportPPLQueryRequest(pplQueryRequest);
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -329,4 +310,36 @@ private Map<String, String> extractFromChatParameters(Map<String, String> parame
         }
         return parameters;
     }
+
+    private String parseOutput(String llmOutput, String indexName) {
+        String ppl;
+        Pattern pattern = Pattern.compile("<ppl>((.|[\\r\\n])+?)</ppl>");
```
zane-neo, line 316: Please add comments and examples to this regex expression.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -329,4 +310,36 @@ private Map<String, String> extractFromChatParameters(Map<String, String> parame
...
+
+                // Modifying the first element
+                if (lists.length > 0) {
+                    lists[0] = "source=" + indexName;
+                }
+
+                // Joining the string back together
+                ppl = String.join("|", lists);
+            } else {
+                ppl = llmOutput;
```
zane-neo, line 337: Is the ppl still correct when it's missing `source=xxx`?

xinyual: No. Should we raise error here? In the logic, it will raise error when executing if we don't start with source=xxx.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Already done/

src/main/java/org/opensearch/agent/tools/PPLTool.java
zane-neo: `Executing` means invoking the ppl transport action, right? If so, I suggest to raise error here since it's not going to success and we can avoid network communication.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -329,4 +310,36 @@ private Map<String, String> extractFromChatParameters(Map<String, String> parame
         }
         return parameters;
     }
+
+    private String parseOutput(String llmOutput, String indexName) {
+        String ppl;
+        Pattern pattern = Pattern.compile("<ppl>((.|[\\r\\n])+?)</ppl>"); // For ppl like <ppl> source=a \n | fields b </ppl>
```
zane-neo, line 316: Can the `\\r\\n` will match `\n`? Do we have UT to ensure this?

xinyual: The logic is directly from here: https://github.com/opensearch-project/dashboards-assistant/blob/e6359634ff6f532afa983ae0785bc0cd53aee425/server/olly/chains/ppl_generator.ts#L247

But you are right, I don't have UT to cover this line. Will add one.

xinyual: Already add. It works.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -254,7 +255,11 @@ private String constructTableInfo(SearchHit[] searchHits, Map<String, MappingMet
     }
 
     private String constructPrompt(String tableInfo, String question, String indexName) {
-        return String.format(Locale.getDefault(), contextPrompt, question.strip(), indexName, tableInfo.strip());
+        Map<String, String> indexInfo = org.opensearch.ml.repackage.com.google.common.collect.ImmutableMap
```
dhrubo-os, line 258: [nit] can't we have that in the import section?

dhrubo-os (approved):

PR #74, opened 2023-12-27
Change js to jar dependency

zane-neo:
### Description
There're two types of dependency in our repo: `on jar` which usually under `org.opensearch`: https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/ and `on plugin` which usually under `org.opensearch.plugin`: https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/. 
For most common cases, we should depend on jars instead of plugins, because depend on plugin is a workaround and depend on jar is the most natural way.

We're depending on sql's plugin because sql doesn't publish jars for `ppl.jar` and `protocol.jar` and sql team refused to publish these jars, so we have to extract them from sql plugin zip.

Job-scheduler publishes jars to this location: `org.opensearch:opensearch-job-scheduler-${version}`: https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/opensearch-job-scheduler/, so we can depend on jar directly. This PR is to make this change.

For Anomaly-detection, the jars also not published to `org.opensearch`: https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/ , and the reason I assume is missing publishAllPublicationsToStagingRepository in scripts/build.sh, for now to not break compilation,
we can only extracting jars from ad plugin. If ad in the future publishes jars to sonatype, we can change ad's dependency to jar's approach just like job-scheduler.

For forced versions: I published ad jars in my local so I can change ad dependency to jar type but found there's jar version conflicts, so forced the versions, after publish the PR, I found ad jars are not in sonatype, so rolled back ad to plugin dependency but didn't remove this force version. I've update this part to remove them.

### Issues Resolved
Change js to jar dependency
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,6 +120,7 @@ dependencies {
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${version}"
+    //JS plugin is published to `org/opensearch` instead of `org/opensearch/plugin` under local maven repo: https://mvnrepository.com/artifact/org.opensearch/opensearch-job-scheduler.
     zipArchive group: 'org.opensearch', name:'opensearch-job-scheduler', version: "${version}"
     zipArchive "org.opensearch.plugin:opensearch-anomaly-detection:${version}"
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-sql-plugin', version: "${version}"
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,6 @@ apply plugin: 'opensearch.testclusters'
 apply plugin: 'opensearch.pluginzip'
 
 def sqlJarDirectory = "$buildDir/dependencies/opensearch-sql-plugin"
-def adJarDirectory = "$buildDir/dependencies/opensearch-time-series-analytics"
 
 configurations {
     zipArchive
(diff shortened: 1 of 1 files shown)


ohltyler:
Could you add more details on the reasoning for this change, and what the blockers are in AD and SQL plugin to do the same? From my understanding it is that AD, for example, doesn't have jars in [sonatype](https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/opensearch-anomaly-detection/3.0.0.0-SNAPSHOT/)? I see it does exist there for [job-scheduler](https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/opensearch-job-scheduler/3.0.0.0-SNAPSHOT/). Do we have a path forward to add?

zane-neo:
> Could you add more details on the reasoning for this change, what it is changing exactly, and what the blockers are in AD and SQL plugin to do the same? From my understanding it is that AD, for example, doesn't have jars in [sonatype](https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/plugin/opensearch-anomaly-detection/3.0.0.0-SNAPSHOT/)? I see it does exist there for [job-scheduler](https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/opensearch-job-scheduler/3.0.0.0-SNAPSHOT/). Do we have a path forward to add?
> 
> Additionally, I see you added forced versions for some dependencies - was this due to jarhell issues when changing the dependency model for JS? I'm worried this will be more of an issue if more plugins with their dependencies are added this way.

@ohltyler Updated the description, please take a look.

ohltyler (approved):
Thanks for the details! Makes sense to me.

PR #78, opened 2023-12-28
Add RAGTool

mingshl:
### Description
Add RAGTool

 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -77,7 +77,7 @@ public abstract class AbstractRetrieverTool implements Tool {
 
     private <T> SearchRequest buildSearchRequest(Map<String, String> parameters) throws IOException {
         String question = parameters.get(INPUT_FIELD);
-        if (StringUtils.isBlank(question) || question == null) {
+        if (StringUtils.isBlank(question)) {
             throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
         }
 
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -77,7 +77,7 @@ public abstract class AbstractRetrieverTool implements Tool {
 
     private <T> SearchRequest buildSearchRequest(Map<String, String> parameters) throws IOException {
         String question = parameters.get(INPUT_FIELD);
-        if (StringUtils.isBlank(question)) {
+        if (StringUtils.isBlank(question) || question == null) {
             throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
         }
 
--- a/src/main/java/org/opensearch/agent/tools/RAGTool.java
+++ b/src/main/java/org/opensearch/agent/tools/RAGTool.java
@@ -110,15 +110,16 @@ public class RAGTool extends AbstractRetrieverTool {
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
         String input = null;
 
+        if (!this.validate(parameters)) {
+            throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
+        }
+
         try {
-            if (!this.validate(parameters)) {
-                throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
-            }
             String question = parameters.get(INPUT_FIELD);
             input = gson.fromJson(question, String.class);
         } catch (Exception e) {
-            log.error("[" + INPUT_FIELD + "] is null or empty, can not process it.", e);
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -77,7 +77,7 @@ public static Map<String, Object> processResponse(SearchHit hit) {
 
     private <T> SearchRequest buildSearchRequest(Map<String, String> parameters) throws IOException {
         String question = parameters.get(INPUT_FIELD);
-        if (StringUtils.isBlank(question)) {
+        if (StringUtils.isBlank(question) || question == null) {
```
ylwu-amzn, line 80: no need to add `|| question == null`

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -0,0 +1,275 @@
...
+        String input = null;
+
+        try {
+            if (!this.validate(parameters)) {
+                throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
+            }
+            String question = parameters.get(INPUT_FIELD);
+            input = gson.fromJson(question, String.class);
+        } catch (Exception e) {
+            log.error("[" + INPUT_FIELD + "] is null or empty, can not process it.", e);
```
ylwu-amzn, line 120: Why this error message same with ling 115 ?  Suggest change to a more general error message as the try block is mainly to catch exception from `gson.fromJson`.

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -0,0 +1,275 @@
...
+    protected String getQueryBody(String queryText) {
+        return queryText;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        String input = null;
+
+        try {
+            if (!this.validate(parameters)) {
```
ylwu-amzn, line 114: Move this `if` block out of `try` block, or change line 115 to `listener.onFailure(new ...); return;`

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
mingshl: removed.

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -0,0 +1,275 @@
...
+                    listener.onResponse((T) modelTensorOutput.getMlModelOutputs());
+                } else {
+                    listener.onResponse((T) outputParser.parse(modelTensorOutput.getMlModelOutputs()));
+                }
+            }, e -> {
+                log.error("Failed to run model " + inferenceModelId, e);
+                listener.onFailure(e);
+            }));
+        }, e -> {
+            log.error("Failed to search index.", e);
```
rbhavna, line 194: Isn't the error here related to failure in running vectorDB tool than in searching the index? Maybe we can change the message?

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: if moving out of the try block, then when the input string is empty, it will throw illegalArgumentException directly.

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: the exception from gson.fromJson is changed to `new IllegalArgumentException("Failed to read question from " + INPUT_FIELD)`

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: good question, it's related to `vectorDBTool.run`, but within `vectorDBTool.run`,  the error is mainly from `client.search`, because it's a recursive call.

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: please see the expected test behavior here after the change for [empty input](https://github.com/opensearch-project/skills/pull/78/files#diff-b17f6e6e035a86ea534f4098c9b59d5df7c74180d6cff353c900d1dfc9e355e1R258), and [malformed input ](https://github.com/opensearch-project/skills/pull/78/files#diff-b17f6e6e035a86ea534f4098c9b59d5df7c74180d6cff353c900d1dfc9e355e1R265)

ylwu-amzn (approved):

ylwu-amzn:
Missed RAGTool in ToolPlugin, check https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/ToolPlugin.java#L63

mingshl:
> https://github.com/opensearch-project/skills/blob/main/src/main/java/org/opensearch/agent/ToolPlugin.java#L63

Created https://github.com/opensearch-project/skills/pull/81 to add tools to ToolPlugin

PR #88, opened 2023-12-29
Integrate search detectors tool with profile API

ohltyler:
### Description
Adds integration with the AD client's new `getDetectorProfile()` method to fetch detailed detector information. It is used in the context of this tool by allowing filtering by detector state - specifically `running`/`disabled`/`failed`. If the user specifies any of these params, the tool will make further transport action calls via `getDetectorProfile()` to fetch the real-time detector state, and include/exclude based on users setting params to `true`/`false`, respectively.

This also unlocks the tool's future capabilities to query things like historical tasks, fine-grained high-cardinality details, initialization progress, error state, etc.

Adds tests to cover the different combinations of state-related params and expected results.

NOTE: the job-scheduler SPI was added as a dependency to resolve classloading issues regarding job-related fields in the `AnomalyDetector` class.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,6 +119,8 @@ dependencies {
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
     compileOnly "org.opensearch:common-utils:${version}"
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
+    compileOnly "org.opensearch:opensearch-job-scheduler-spi:${version}"
+
 
     // ZipArchive dependencies used for integration tests
     zipArchive group: 'org.opensearch.plugin', name:'opensearch-ml-plugin', version: "${version}"
--- a/src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
@@ -138,9 +138,6 @@ public class SearchAnomalyDetectorsTool implements Tool {
 
             // If we need to filter by detector state, make subsequent profile API calls to each detector
             if (running != null || disabled != null || failed != null) {
-
-                // Send out individual AD client calls to fetch detector profiles, continuously adding to a
-                // tracked list of CompletableFutures
                 List<CompletableFuture<GetAnomalyDetectorResponse>> profileFutures = new ArrayList<>();
                 for (SearchHit hit : hits) {
                     CompletableFuture<GetAnomalyDetectorResponse> profileFuture = new CompletableFuture<>();
@@ -166,9 +163,6 @@ public class SearchAnomalyDetectorsTool implements Tool {
                     );
                     adClient.getDetectorProfile(profileRequest, profileListener);
                 }
-
-                // Wait for all CompletableFutures to complete, and iterate through the responses. Filter out
-                // detectors with unwanted detector states.
                 CompletableFuture<List<GetAnomalyDetectorResponse>> listFuture = CompletableFuture
                     .allOf(profileFutures.toArray(new CompletableFuture<?>[0]))
                     .thenApply(v -> profileFutures.stream().map(CompletableFuture::join).collect(Collectors.toList()));
@@ -194,7 +188,7 @@ public class SearchAnomalyDetectorsTool implements Tool {
                                 detectorState = DetectorStateString.Failed.name();
                             } else {
                                 // Task states may fall under other values, such as "FEATURE_REQUIRED" / "STOPPED" / etc.
-                                // We assume here that these will all fall under the disabled category
+                                // Defaulting all other states to fall under the disabled category
                                 detectorState = DetectorStateString.Disabled.name();
                             }
                         } else {
--- a/src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
@@ -10,6 +10,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
@@ -140,7 +141,8 @@ public class SearchAnomalyDetectorsTool implements Tool {
             if (running != null || disabled != null || failed != null) {
                 List<CompletableFuture<GetAnomalyDetectorResponse>> profileFutures = new ArrayList<>();
                 for (SearchHit hit : hits) {
-                    CompletableFuture<GetAnomalyDetectorResponse> profileFuture = new CompletableFuture<>();
+                    CompletableFuture<GetAnomalyDetectorResponse> profileFuture = new CompletableFuture<GetAnomalyDetectorResponse>()
+                        .orTimeout(30000, TimeUnit.MILLISECONDS);
                     profileFutures.add(profileFuture);
                     ActionListener<GetAnomalyDetectorResponse> profileListener = ActionListener
                         .<GetAnomalyDetectorResponse>wrap(profileResponse -> {
@@ -163,17 +165,25 @@ public class SearchAnomalyDetectorsTool implements Tool {
                     );
                     adClient.getDetectorProfile(profileRequest, profileListener);
                 }
-                CompletableFuture<List<GetAnomalyDetectorResponse>> listFuture = CompletableFuture
-                    .allOf(profileFutures.toArray(new CompletableFuture<?>[0]))
-                    .thenApply(v -> profileFutures.stream().map(CompletableFuture::join).collect(Collectors.toList()));
-                final List<GetAnomalyDetectorResponse> profileResponses = listFuture.join();
+
+                List<GetAnomalyDetectorResponse> profileResponses = new ArrayList<>();
+                try {
+                    CompletableFuture<List<GetAnomalyDetectorResponse>> listFuture = CompletableFuture
+                        .allOf(profileFutures.toArray(new CompletableFuture<?>[0]))
+                        .thenApply(v -> profileFutures.stream().map(CompletableFuture::join).collect(Collectors.toList()));
+                    profileResponses = listFuture.join();
+                } catch (Exception e) {
+                    log.error("Failed to get all anomaly detector profiles.", e);
+                    listener.onFailure(e);
+                }
+
                 for (GetAnomalyDetectorResponse profileResponse : profileResponses) {
                     if (profileResponse != null && profileResponse.getDetector() != null) {
                         String detectorId = profileResponse.getDetector().getId();
 
                         // We follow the existing logic as the frontend to determine overall detector state
                         // https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/server/routes/utils/adHelpers.ts#L437
-                        String detectorState;
+                        String detectorState = DetectorStateString.Disabled.name();
                         ADTask realtimeTask = profileResponse.getRealtimeAdTask();
 
                         if (realtimeTask != null) {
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                        hit.getId(),
+                        Versions.MATCH_ANY,
+                        true,
+                        true,
+                        "",
+                        "",
+                        false,
+                        null
+                    );
+                    adClient.getDetectorProfile(profileRequest, profileListener);
```
jackiehanyang, line 164: can we move this `getDetectorProfile` call inside a try catch block?

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+
+                        if (running != null && !running && detectorState.equals(DetectorStateString.Running.name())) {
+                            hitsAsMap.remove(detectorId);
+                        }
+                        if (disabled != null && !disabled && detectorState.equals(DetectorStateString.Disabled.name())) {
+                            hitsAsMap.remove(detectorId);
+                        }
+                        if (failed != null && !failed && detectorState.equals(DetectorStateString.Failed.name())) {
+                            hitsAsMap.remove(detectorId);
+                        }
```
jackiehanyang, line 206: It seems like we're doing a similar check and action for different detector states? Could we reduces code duplication and centralizes the filtering logic here?

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
             StringBuilder sb = new StringBuilder();
-            SearchHit[] hits = response.getHits().getHits();
+            List<SearchHit> hits = Arrays.asList(response.getHits().getHits());
+            Map<String, SearchHit> hitsAsMap = hits.stream().collect(Collectors.toMap(SearchHit::getId, hit -> hit));
+
+            // If we need to filter by detector state, make subsequent profile API calls to each detector
+            if (running != null || disabled != null || failed != null) {
+                List<CompletableFuture<GetAnomalyDetectorResponse>> profileFutures = new ArrayList<>();
+                for (SearchHit hit : hits) {
+                    CompletableFuture<GetAnomalyDetectorResponse> profileFuture = new CompletableFuture<>();
```
jackiehanyang, line 143: can `hits` be vary large? should we add a timeout to the `CompletableFuture` here to avoid potential resource exhaustion?

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: The logic itself can't be simplified since they have to be grouped as they currently are, but I will combine into one conditional.

jackiehanyang: maybe something like this:
```suggestion
      private void filterDetectorsByState(Map<String, SearchHit> hitsAsMap, 
                                                   Boolean stateCondition, 
                                                   String targetState, 
                                                   String actualState) {
    if (stateCondition != null && !stateCondition && actualState.equals(targetState)) {
        hitsAsMap.remove(detectorId);
    }
}

filterDetectorsByState(hitsAsMap, running, DetectorStateString.Running.name(), detectorState);
filterDetectorsByState(hitsAsMap, disabled, DetectorStateString.Disabled.name(), detectorState);
filterDetectorsByState(hitsAsMap, failed, DetectorStateString.Failed.name(), detectorState);
```

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                        "",
+                        false,
+                        null
+                    );
+                    adClient.getDetectorProfile(profileRequest, profileListener);
+                }
+                CompletableFuture<List<GetAnomalyDetectorResponse>> listFuture = CompletableFuture
+                    .allOf(profileFutures.toArray(new CompletableFuture<?>[0]))
+                    .thenApply(v -> profileFutures.stream().map(CompletableFuture::join).collect(Collectors.toList()));
+                final List<GetAnomalyDetectorResponse> profileResponses = listFuture.join();
```
dbwiddis, line 169: This throws some exceptions and needs to be inside a catch block, or will get silently swallowed in this lambda.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                for (GetAnomalyDetectorResponse profileResponse : profileResponses) {
+                    if (profileResponse != null && profileResponse.getDetector() != null) {
+                        String detectorId = profileResponse.getDetector().getId();
+
+                        // We follow the existing logic as the frontend to determine overall detector state
+                        // https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/server/routes/utils/adHelpers.ts#L437
+                        String detectorState;
```
dbwiddis, line 176: (Style) Suggestion: iniitalize here to `DetectorStateString.Disabled.name()` and then you don't have multiple `else` blocks setting it that way.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: I like this! Thanks for the suggestion, I think this is the most readable and flexible.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                            } else {
+                                // Task states may fall under other values, such as "FEATURE_REQUIRED" / "STOPPED" / etc.
+                                // Defaulting all other states to fall under the disabled category
+                                detectorState = DetectorStateString.Disabled.name();
+                            }
+                        } else {
+                            detectorState = DetectorStateString.Disabled.name();
+                        }
+
+                        if (running != null && !running && detectorState.equals(DetectorStateString.Running.name())) {
```
dbwiddis, line 198: (Style) Here and two below.  If you use `Boolean.FALSE.equals(running)` it handles the null check.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
dbwiddis: Could probably do the map removal as part of the conditional above which sets `detectorState`.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
```diff
@@ -123,24 +131,92 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
-            for (SearchHit hit : hits) {
+            for (SearchHit hit : hitsAsMap.values()) {
                 sb.append("{");
                 sb.append("id=").append(hit.getId()).append(",");
                 sb.append("name=").append(hit.getSourceAsMap().get("name"));
                 sb.append("}");
             }
             sb.append("]");
-            sb.append("TotalAnomalyDetectors=").append(response.getHits().getTotalHits().value);
+            sb.append("TotalAnomalyDetectors=").append(hitsAsMap.values().size());
```
dbwiddis, line 219:
```suggestion
            sb.append("TotalAnomalyDetectors=").append(hitsAsMap.size());
```

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: This won't throw anything since its a wrapper around `client.execute()` that is void since it is just sending out an async request. The error handling is done for these profile calls is on lines 149-151 if the `profileListener` returns `onFailure()`.

dbwiddis (commented):
Needs a try-catch, at least. Some other suggestions.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Yes it could potentially be large if the user has hundreds of detectors, overrides the default of 20, and asks a broad question that would return some/all of them. Good callout on adding a timeout, I will add.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Given @dbwiddis suggestion already removed one of the conditionals, I'm going to just combine into a single conditional to simplify things instead of adding a helper fn in this case. This number of states shouldn't change or scale out over time.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Cool! thanks for the suggestion.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: I could integrate the map removal as part of the above conditionals, but I prefer to leave as-is for more readability. I would need to leave in the extra else blocks as well.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Added 30s timeout.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Added

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Good idea - added.

src/main/java/org/opensearch/agent/tools/SearchAnomalyDetectorsTool.java
ohltyler: Added

build.gradle
```diff
@@ -119,6 +119,8 @@ dependencies {
     implementation fileTree(dir: sqlJarDirectory, include: ["opensearch-sql-${version}.jar", "ppl-${version}.jar", "protocol-${version}.jar"])
     compileOnly "org.opensearch:common-utils:${version}"
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
+    compileOnly "org.opensearch:opensearch-job-scheduler-spi:${version}"
```
Hailong-am, line 122: If it's `compileOnly`, who will provide the class during runtime? I don't think skills depends on job scheduler

ohltyler: This is needed for AD tool due to AD's dependency. I've used the same logic as AD plugin here for consistency regarding the job scheduler dependencies.

jackiehanyang (approved):

dbwiddis (approved):
LGTM, thanks for the changes.

PR #93, opened 2024-01-02
fix: some parsing exception from search index tool.

yuye-aws:
### Description
This PR can fix some parsing exceptions in search index tool.
 
### Issues Resolved
The LLM returns input parameters for search index tool in `action_input`. Before executing the DSL query to search the index, we need to parse the `query` field from `action_input`. Sometimes the DSL query given by LLM is problematic and thus lead to parsing exception. To support more user cases, we add two parsing methods in this PR to make it more likely to successfully execute search index tool.

This PR also adds one more unit test case for model group index.
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
@@ -80,12 +80,11 @@ public class SearchIndexTool implements Tool {
 
     private SearchRequest getSearchRequest(String index, String query) throws IOException {
         SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
-        XContentParser queryParser = XContentType.JSON
-                .xContent()
-                .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, query);
+        XContentParser queryParser = XContentType.JSON.xContent().createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, query);
         searchSourceBuilder.parseXContent(queryParser);
         return new SearchRequest().source(searchSourceBuilder).indices(index);
     }
+
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
         try {
@@ -102,8 +101,7 @@ public class SearchIndexTool implements Tool {
                     // try different json parsing method
                     query = jsonObject.get(QUERY_FIELD).getAsString();
                     searchRequest = getSearchRequest(index, query);
-                }
-                catch (Exception e2) {
+                } catch (Exception e2) {
                     // try wrapped query
                     query = "{\"query\": " + query + "}";
                     searchRequest = getSearchRequest(index, query);
--- a/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
@@ -7,6 +7,7 @@ package org.opensearch.agent.tools;
 
 import static org.opensearch.ml.common.CommonValue.*;
 
+import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


xinyual (approved):

xinyual (approved):

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -77,21 +78,35 @@ public boolean validate(Map<String, String> parameters) {
...
+        return new SearchRequest().source(searchSourceBuilder).indices(index);
+    }
+
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
         try {
             String input = parameters.get(INPUT_FIELD);
             JsonObject jsonObject = StringUtils.gson.fromJson(input, JsonObject.class);
             String index = jsonObject.get(INDEX_FIELD).getAsString();
             String query = jsonObject.get(QUERY_FIELD).toString();
```
zane-neo, line 94: This line is not necessary, can be removed.

zane-neo, line 94: This line is not necessary, can be removed.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -77,21 +78,35 @@ public boolean validate(Map<String, String> parameters) {
...
+            try {
+                searchRequest = getSearchRequest(index, query);
+            } catch (Exception e1) {
+                try {
+                    // try different json parsing method
+                    query = jsonObject.get(QUERY_FIELD).getAsString();
+                    searchRequest = getSearchRequest(index, query);
+                } catch (Exception e2) {
+                    // try wrapped query
+                    query = "{\"query\": " + query + "}";
```
zane-neo, line 106: Based on the discussion, this tool only accept index name and correct query DSL, no necessary to manually rectify the DSL.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -77,21 +78,35 @@ public boolean validate(Map<String, String> parameters) {
...
-            searchSourceBuilder.parseXContent(queryParser);
-            SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(index);
+
+            SearchRequest searchRequest;
+            try {
+                searchRequest = getSearchRequest(index, query);
+            } catch (Exception e1) {
+                try {
+                    // try different json parsing method
+                    query = jsonObject.get(QUERY_FIELD).getAsString();
```
zane-neo, line 102: Any test cover this line? Is it safe to invoke `getAsString()` on an jsonElement?

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -77,21 +78,35 @@ public boolean validate(Map<String, String> parameters) {
...
-                .xContent()
-                .createParser(xContentRegistry, LoggingDeprecationHandler.INSTANCE, query);
-            searchSourceBuilder.parseXContent(queryParser);
-            SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(index);
+
+            SearchRequest searchRequest;
+            try {
+                searchRequest = getSearchRequest(index, query);
+            } catch (Exception e1) {
+                try {
```
zane-neo, line 100: We can remove everything under catch and simply add error logs here.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
yuye-aws: Got it.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
yuye-aws: There is a test case named `testRunWithEmptyQueryBody` to cover this line. Sometimes the query is `"{}"`. If the previous parsing method fails, we can try parse with `getAsString()`. If there is any exception, this tool will call `listener.onFailure(e);`

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
yuye-aws: I think we should at least try two parsing methods: `jsonObject.get(QUERY_FIELD).toString();` and `jsonObject.get(QUERY_FIELD).getAsString();`. If both methods fail, we need to show the error log.

PR #94, opened 2024-01-02
fix: read http entity only once in integ test

zhichao-aws:
### Description
In main and 2.x we use different http dependency in integ test. In main the dependency is from org.apache.hc.core5.http, and in 2.x it's from org.apache.http.
Current integ test code in main works well, but in 2.x, we can not read a http entity twice. The entity is a stream, if read twice, the result is null. We can solve this by cache the result and read it only once https://github.com/opensearch-project/skills/pull/91/commits/28bfc01f703912d12c24ba138e846c34aeeb8bde.
Although in main it works well, we can also modify this to keep it consistent between main and 2.x. If we modify some code in the future, this issue will not be a blocker for auto backport.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/test/java/org/opensearch/integTest/BaseAgentToolsIT.java
+++ b/src/test/java/org/opensearch/integTest/BaseAgentToolsIT.java
@@ -75,10 +75,9 @@ public abstract class BaseAgentToolsIT extends OpenSearchSecureRestTestCase {
     }
 
     @SneakyThrows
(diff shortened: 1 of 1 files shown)


zhichao-aws:
No need to backport, the fix is already in 2.x

zane-neo (approved):

PR #98, opened 2024-01-03
provide PPL for querying system indices

xinyual:
### Description
provide PPL for querying system indices
 
### Issues Resolved
provide PPL for querying system indices. Will address recursively replacement.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.agent.tools;
 
+import static org.opensearch.ml.common.CommonValue.*;
+
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
@@ -63,6 +65,9 @@ public class PPLTool implements Tool {
 
     public static final String TYPE = "PPLTool";
 
+    public static final List<String> SYSTEM_INDICES = List
+        .of(ML_MODEL_GROUP_INDEX, ML_MODEL_INDEX, ML_TASK_INDEX, ML_CONNECTOR_INDEX, ML_CONFIG_INDEX, ML_MODEL_CONTROLLER_INDEX);
+
     @Setter
     private Client client;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -63,6 +65,9 @@ public class PPLTool implements Tool {
 
     public static final String TYPE = "PPLTool";
 
+    public static final List<String> SYSTEM_INDICES = List
+        .of(ML_MODEL_GROUP_INDEX, ML_MODEL_INDEX, ML_TASK_INDEX, ML_CONNECTOR_INDEX, ML_CONFIG_INDEX, ML_MODEL_CONTROLLER_INDEX);
```
zane-neo, line 69: There could be more indices like: agent, memory etc, using regex like `.plugins-ml-*` is more flexible to adapt to future changes. BTW, do we need to consider other system indices which are not ml-common's?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -97,6 +102,11 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
         if (StringUtils.isBlank(indexName) || StringUtils.isBlank(question)) {
             throw new IllegalArgumentException("Parameter index and question can not be null or empty.");
         }
+        for (String systemIndex : SYSTEM_INDICES) {
+            if (indexName.contains(systemIndex)) {
+                throw new IllegalArgumentException("We cannot search system indices " + indexName);
```
zane-neo, line 107:
```suggestion
                throw new IllegalArgumentException("PPLTool doesn't support searching system indices, current searching index name: " + indexName);
```

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: I think so. Is there a pattern to match all system indices?

zane-neo: Looks like all system indices are starting with `.`, here are some examples: `.opendistro-alerting-config`,`.opensearch-observability`, `.replication-metadata-store`. Check if starting with `.` could be safe and also fine-tune the error message.

xinyual: Already done.

zane-neo (approved):

PR #100, opened 2024-01-03
fix: update search index tool parse logic

yuye-aws:
### Description
This PR resolves some extra code review comments in https://github.com/opensearch-project/skills/pull/93.
 
### Issues Resolved
This PR resolves some extra code review comments in https://github.com/opensearch-project/skills/pull/93. Some  unit tests are also updated accordingly
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
@@ -90,8 +90,12 @@ public class SearchIndexTool implements Tool {
         try {
             String input = parameters.get(INPUT_FIELD);
             JsonObject jsonObject = StringUtils.gson.fromJson(input, JsonObject.class);
-            String index = jsonObject.get(INDEX_FIELD).getAsString();
-            String query = jsonObject.get(QUERY_FIELD).toString();
+            String index = Optional.ofNullable(jsonObject).map(x -> x.get(INDEX_FIELD)).map(JsonElement::getAsString).orElse(null);
+            String query = Optional.ofNullable(jsonObject).map(x -> x.get(QUERY_FIELD)).map(JsonElement::toString).orElse(null);
+            if (index == null || query == null) {
+                listener.onFailure(new IllegalArgumentException("SearchIndexTool's two parameter: index and query are required!"));
+                return;
+            }
             SearchRequest searchRequest = getSearchRequest(index, query);
 
             ActionListener<SearchResponse> actionListener = ActionListener.<SearchResponse>wrap(r -> {
--- a/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
@@ -12,6 +12,7 @@ import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
@@ -30,6 +31,7 @@ import org.opensearch.ml.common.utils.StringUtils;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.builder.SearchSourceBuilder;
 
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import lombok.Getter;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -48,7 +48,7 @@ public class SearchIndexTool implements Tool {
 
     public static final String TYPE = "SearchIndexTool";
     private static final String DEFAULT_DESCRIPTION =
-        "Use this tool to search an index by providing two parameters: 'index' for the index name, and 'query' for the OpenSearch DSL formatted query.";
+        "Use this tool to search an index by providing two parameters: 'index' for the index name, and 'query' for the OpenSearch DSL formatted query. Only use this tool when a DSL query is available.";
```
zane-neo, line 51: Have you tested with this description, will LLM still generates DSL automatically and then invoke this tool? Or this tool will only be invoked when user DSL is provided?

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -97,15 +97,9 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
-                    query = jsonObject.get(QUERY_FIELD).getAsString();
-                    searchRequest = getSearchRequest(index, query);
-                } catch (Exception e2) {
-                    // try wrapped query
-                    query = "{\"query\": " + query + "}";
-                    searchRequest = getSearchRequest(index, query);
-                }
+                // try different json parsing method
+                query = jsonObject.get(QUERY_FIELD).getAsString();
+                searchRequest = getSearchRequest(index, query);
```
zane-neo, line 102: Remove this since it would cause exception when `jsonObject.get(QUERY_FIELD)` returns jsonElement. And add error log here.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -97,15 +97,9 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             try {
                 searchRequest = getSearchRequest(index, query);
             } catch (Exception e1) {
```
zane-neo, line 99:
```suggestion
            } catch (Exception e) {
```

src/test/java/org/opensearch/agent/tools/SearchIndexToolTests.java
```diff
@@ -85,7 +85,7 @@ public void testValidateWithEmptyInput() {
 
     @Test
     public void testRunWithNormalIndex() {
-        String inputString = "{\"index\": \"test-index\", \"query\": {\"query\": {\"match_all\": {}}}}";
+        String inputString = "{\"index\": \"test-index\", \"query\": \"{}\"}";
```
zane-neo, line 88: This is not a correct DSL.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -96,16 +96,10 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             SearchRequest searchRequest;
             try {
```
zane-neo, line 97: Remove this try...catch block since the outer try...catch can catch the exception here.

yuye-aws: This try catch block is needed because we need to try another different parsing method.

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
yuye-aws: I have tested 8 user cases from tool selection. None of the case will trigger search index tool.

src/test/java/org/opensearch/agent/tools/SearchIndexToolTests.java
yuye-aws: I will update this with a normal DSL

src/main/java/org/opensearch/agent/tools/SearchIndexTool.java
```diff
@@ -92,21 +92,7 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             JsonObject jsonObject = StringUtils.gson.fromJson(input, JsonObject.class);
             String index = jsonObject.get(INDEX_FIELD).getAsString();
             String query = jsonObject.get(QUERY_FIELD).toString();
```
zane-neo, line 94:
```suggestion
            String index = Optional.ofNullable(jsonObject).map(x -> x.get(INDEX_FIELD)).map(JsonElement::getAsString).orElse(null);
            String query = Optional.ofNullable(jsonObject).map(x -> x.get(QUERY_FIELD)).map(JsonElement::toString).orElse(null);
            if (index == null || query == null) {
                listener.onFailure(new IllegalArgumentException("SearchIndexTool's two parameter: index and query are required!"));
                return;
            }
```

zane-neo (approved):

PR #101, opened 2024-01-03
ignore alias field

xinyual:
### Description
Ignore Alias field in the table information. Otherwise, it will confuse the LLM. The alias field cannot be executed by PPL.
 
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -285,7 +285,9 @@ public class PPLTool implements Tool {
             if (v instanceof Map) {
                 Map<String, Object> vMap = (Map<String, Object>) v;
                 if (vMap.containsKey("type")) {
-                    fieldsToType.put(prefix + n, (String) vMap.get("type"));
+                    if (!((vMap.get("type")).equals("alias"))) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -285,7 +285,9 @@ private void extractNamesTypes(Map<String, Object> mappingSource, Map<String, St
             if (v instanceof Map) {
                 Map<String, Object> vMap = (Map<String, Object>) v;
                 if (vMap.containsKey("type")) {
-                    fieldsToType.put(prefix + n, (String) vMap.get("type"));
+                    if (!((vMap.get("type")).equals("alias"))) {
```
zane-neo, line 288: Is it possible that vMap.get("type") returns null?

xinyual: I don't think so. Every field inside our mapping should have a field. But I change it to getOrDefault for safety.

zane-neo (approved):

zane-neo:
Code coverage dropped a lot, seems more UTs are needed.

xinyual:
Will add UT in a new PR

PR #111, opened 2024-01-08
Set default index pattern for search AD results tool

ohltyler:
### Description
There was a bug where the default `indices` field for the `SearchRequest` passed to the AD client was not set, causing an error in the respective AD transport action. This fixes that bug by adding a default index pattern. In the future we may support propagating custom result indices to the tool.

Added corresponding UT.

 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
@@ -13,6 +13,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.ad.client.AnomalyDetectionNodeClient;
+import org.opensearch.agent.tools.utils.ToolConstants;
 import org.opensearch.client.Client;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.index.query.BoolQueryBuilder;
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
```diff
@@ -130,7 +130,9 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             .from(startIndex)
             .sort(sortString, sortOrder);
 
-        SearchRequest searchAnomalyResultsRequest = new SearchRequest().source(searchSourceBuilder);
+        // In the future we may support custom result indices. For now default to the default AD result system indices.
+        String resultIndices = ".opendistro-anomaly-results*";
```
owaiskazi19, line 134: Do we have any file that stores constant?

owaiskazi19 (approved):
One small comment. Rest LGTM

src/main/java/org/opensearch/agent/tools/SearchAnomalyResultsTool.java
ohltyler: Nothing clean exposed from AD plugin. Considered this a one-off so just hardcoded.

ohltyler: But good point, I will add to `ToolConstants` file.

ohltyler: Updated

owaiskazi19 (approved):

PR #125, opened 2024-01-12
Add default prompt to ppl tool

xinyual:
### Description
add default prompt to ppl tool
 
### Issues Resolved
add default prompt to ppl tool. So customer don't need to add very low context prompt in their request body.
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -373,4 +405,20 @@ private String parseOutput(String llmOutput, String indexName) {
         return ppl;
     }
 
+    private String loadDefaultPrompt(PPLModelType pplModelType) {
```
zhichao-aws, line 408: Now we read the resource every time when invoke creating PPLTool. It would be more efficient if we only read them once and store the prompts at static fields and

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -373,4 +405,20 @@ private String parseOutput(String llmOutput, String indexName) {
...
+        try (InputStream searchResponseIns = PPLTool.class.getResourceAsStream("PPLDefaultPrompt.json");) {
+            if (searchResponseIns != null) {
+                String defaultPromptContent = new String(searchResponseIns.readAllBytes(), StandardCharsets.UTF_8);
+                Map<String, String> defaultPromptDict = gson.fromJson(defaultPromptContent, Map.class);
+                return defaultPromptDict.get(pplModelType.toString());
```
zhichao-aws, line 413: What if pplModelType is empty string ("")? This happens when user gives neither prompt and model_type

xinyual: The default value for model type is claude and will use default prompt for claude if both fields are none.

zhichao-aws: got it

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Already move it to static.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -83,12 +87,46 @@ public class PPLTool implements Tool {
...
     private static Gson gson = new Gson();
 
-    public PPLTool(Client client, String modelId, String contextPrompt) {
+    private static Map<String, String> defaultPromptDict;
+
+    static {
+        try {
+            defaultPromptDict = loadDefaultPromptDict();
+        } catch (IOException e) {
+            throw new RuntimeException("fail to load default prompt dict" + e.getMessage());
```
zane-neo, line 100: Is it possible that this would break the plugin initialization process with this error thrown?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -83,12 +87,46 @@ public class PPLTool implements Tool {
...
+
+    }
+
+    public PPLTool(Client client, String modelId, String contextPrompt, String pplModelType) {
         this.client = client;
         this.modelId = modelId;
-        this.contextPrompt = contextPrompt;
+        this.pplModelType = PPLModelType.from(pplModelType);
+        if (contextPrompt.isEmpty()) {
+            this.contextPrompt = this.defaultPromptDict.getOrDefault(this.pplModelType.toString(), "");
```
zane-neo, line 126: Is it possible this pplModelType.toString() produce NPE?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -83,12 +87,46 @@ public class PPLTool implements Tool {
 
     private String contextPrompt;
 
+    private PPLModelType pplModelType;
+
     private static Gson gson = new Gson();
 
-    public PPLTool(Client client, String modelId, String contextPrompt) {
+    private static Map<String, String> defaultPromptDict;
```
zane-neo, line 94: Is there other tools would use default prompt? If so, it's better to create an abstract method to reduce the code duplication in other tools.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Since we init the modeltype with the default value, the model type should be an object or throw an error. So it shouldn't be a NPE.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: I think PPL would be the current only one.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: You're right. Already change the logic to init the dict to an empty dict and log an error here.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -83,12 +87,47 @@ public class PPLTool implements Tool {
...
+        FINETUNE;
+
+        public static PPLModelType from(String value) {
+            if (value.isEmpty()) {
+                return PPLModelType.CLAUDE;
+            }
+            try {
+                return PPLModelType.valueOf(value.toUpperCase(Locale.ROOT));
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Wrong PPL Model type, should be CLAUDE or FINETUNE");
```
zane-neo, line 116: Is this possible to break the initialization flow?

xinyual: Yes, if the customer provide the wrong model type. Do you think it's better to set the default model type claude always?

zane-neo: Yes, let's do that.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Find Index routing tool would be another one. So add the abstractPromptedTool. Please check it.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Already add the logic.

xinyual:
Force push since we decide not to add abstract prompt class this time.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -83,12 +87,48 @@ public class PPLTool implements Tool {
...
+            defaultPromptDict = new HashMap<>();
+        }
+    }
+
+    public enum PPLModelType {
+        CLAUDE,
+        FINETUNE;
+
+        public static PPLModelType from(String value) {
+            if (value.isEmpty()) {
```
zane-neo, line 110: Is value possibly be null?

xinyual: No, when create ppltool object, we use `(String) map.getOrDefault("model_type", "")` So model type string will have default value.

src/main/java/org/opensearch/agent/tools/PPLTool.java
zane-neo: Ok, but there's no need to invoke toString() on a String class, let's remove this unnecessary call.

zane-neo (approved):

PR #131, opened 2024-01-16
add previous key for PPL tool

xinyual:
### Description
add previous key for PPL tool, so we can extract content from previous tool output.
request body would be like:
```
{
      "type": "CatIndexTool",
      "name": "CatIndexTool2",
      "description": "Use this tool to get OpenSearch index information: (health, status, index, uuid, primary count, replica count, docs.count, docs.deleted, store.size, primary.store.size). Also, use this to get all index name. You should always use this tool when you are asked questions about indices number."
    },
      {
      "type": "PPLTool",
      "name": "TransferQuestionToPPLAndExecuteTool",
      "description": "Use this tool to transfer natural language to generate PPL and execute PPL to query inside. Use this tool after you know the index name, otherwise, call IndexRoutingTool first. The input parameters are: {index:IndexName, question:UserQuestion}",
      "parameters": {
        "model_id": "PrlMEI0BavJyXuD79Loi",
        "model_type": "CLAUDE",
        "previous_tool_name": "CatIndexTool3",
        "response_filter": "$.completion"
      }
    }
 
```
The previous_tool_name should be name of previous tool
### Issues Resolved
add previous key for PPL tool, so we can extract content from previous tool output.
 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -129,12 +131,20 @@ public PPLTool(Client client, String modelId, String contextPrompt, String pplMo
...
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
         parameters = extractFromChatParameters(parameters);
-        String indexName = parameters.get("index");
+        String indexName;
+        if (parameters.containsKey("index")) {
+            indexName = parameters.get("index");
+        } else if (!StringUtils.isBlank(this.previousToolKey) && parameters.containsKey(this.previousToolKey + ".output")) {
+            indexName = parameters.get(this.previousToolKey + ".output"); // read index name from previous key
```
Hailong-am, line 144: should we handle 'Not sure' here ?

xinyual: Ok. Once the index routing pr is merged I will check whether the index is equal to NOT_SURE and then throw an exception if it is.

zane-neo:
This should work for PPLTool case, but I'm thinking of creating a systematically approach to deal with the tool results passing through the agent execute flow, e.g. a map contains all tool's result with iteration number. That would need bigger effort but that should be the correct direction. Please evaluate the effort and better creating a design doc to involve broader audience.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -213,7 +222,17 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             ));
         }, e -> {
             log.info("fail to get mapping: " + e);
-            listener.onFailure(e);
+            String error_message = e.getMessage();
```
zane-neo, line 225: please use camel naming convention for variables

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -431,6 +451,20 @@ private String parseOutput(String llmOutput, String indexName) {
         return ppl;
     }
 
+    private String getIndexNameFromParameters(Map<String, String> parameters) {
+        String indexName = "";
+        if (parameters.containsKey("index")) {
```
zane-neo, line 456: this check is redundant since the next line is using `getOrDefault`

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -431,6 +451,20 @@ private String parseOutput(String llmOutput, String indexName) {
...
+            indexName = parameters.getOrDefault("index", "");
+        }
+        if (!StringUtils.isBlank(this.previousToolKey)
+            && parameters.containsKey(this.previousToolKey + ".output")
```
zane-neo, line 460: same above, this containsKey is redundant.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -213,7 +222,17 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             ));
         }, e -> {
             log.info("fail to get mapping: " + e);
-            listener.onFailure(e);
+            String error_message = e.getMessage();
+            if (error_message.contains("no such index")) {
+                listener
+                    .onFailure(
+                        new IllegalArgumentException(
+                            "Return this final answer to human directly and do not use other tools: 'Please provide index name'. Please try to directly send this message to human to ask for index name"
```
zane-neo, line 230: Should we change the message to something like: `the index is not exist, please provide a correct index name ...`

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -431,6 +451,16 @@ private String parseOutput(String llmOutput, String indexName) {
         return ppl;
     }
 
+    private String getIndexNameFromParameters(Map<String, String> parameters) {
+        String indexName = "";
+        indexName = parameters.getOrDefault("index", "");
```
zane-neo, line 456: This two line can be merged together

xinyual: Already changed

zane-neo (approved):

PR #139, opened 2024-01-22
fix name bug

xinyual:
### Description
When we use pattern, the return metadata map is <index_name>: <index_mapping>. So we need to use the name from mapping instead of pattern name.
 

 
### Check List
  - [ ] All tests pass


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -149,9 +149,12 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
         GetMappingsRequest getMappingsRequest = buildGetMappingRequest(indexName);
         client.admin().indices().getMappings(getMappingsRequest, ActionListener.<GetMappingsResponse>wrap(getMappingsResponse -> {
             Map<String, MappingMetadata> mappings = getMappingsResponse.getMappings();
+            if (mappings.size() == 0) {
+                throw new IllegalArgumentException("No matching index with index name: " + indexName);
```
zane-neo, line 153: Seems we should use`No matching mapping with ...`, is this index from user's input?

zane-neo (approved):

PR #140, opened 2024-01-22
Enhance RagTool to choose neural sparse query type

mingshl:
### Description
semantic search is based on text embedding models,  and neural sparse search is based on sparse embedding models. We are now enhancing RagTool with a parameter `query_type`, to enable both options. 

now that users can set parameter `query_type=neural` to run neural query , 
or  set parameter `query_type=neural_sparse` to run neural_sparse query  
 
 
There are two phases of RAGTool, first is through retrieval phase, to query, and second phase is content generation through inference mode. 

The newly added parameter `enable_Content_Generation`, the default is true, When it sets to false, it will return search retrieval result directly.  When it sets to true, the RAGTool will continue trigger inference model to generate content. 

Also fixing the build failure related to JDK 21 

### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/RAGTool.java
+++ b/src/main/java/org/opensearch/agent/tools/RAGTool.java
@@ -6,7 +6,7 @@
 package org.opensearch.agent.tools;
 
 import static org.apache.commons.lang3.StringEscapeUtils.escapeJson;
-import static org.opensearch.agent.tools.VectorDBTool.DEFAULT_K;
+import static org.opensearch.agent.tools.AbstractRetrieverTool.*;
 import static org.opensearch.ml.common.utils.StringUtils.gson;
 import static org.opensearch.ml.common.utils.StringUtils.toJson;
 
@@ -21,10 +21,10 @@ import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
-import org.opensearch.ml.common.output.model.ModelTensor;
 import org.opensearch.ml.common.output.model.ModelTensorOutput;
 import org.opensearch.ml.common.output.model.ModelTensors;
 import org.opensearch.ml.common.spi.tools.Parser;
+import org.opensearch.ml.common.spi.tools.Tool;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskRequest;
@@ -44,29 +44,28 @@ import lombok.extern.log4j.Log4j2;
 @Setter
 @Getter
 @ToolAnnotation(RAGTool.TYPE)
-public class RAGTool extends AbstractRetrieverTool {
+public class RAGTool implements Tool {
     public static final String TYPE = "RAGTool";
     public static String DEFAULT_DESCRIPTION =
         "Use this tool to retrieve helpful information to optimize the output of the large language model to answer questions.";
     public static final String INFERENCE_MODEL_ID_FIELD = "inference_model_id";
     public static final String EMBEDDING_MODEL_ID_FIELD = "embedding_model_id";
+    public static final String INDEX_FIELD = "index";
+    public static final String SOURCE_FIELD = "source_field";
+    public static final String DOC_SIZE_FIELD = "doc_size";
     public static final String EMBEDDING_FIELD = "embedding_field";
     public static final String OUTPUT_FIELD = "output_field";
     public static final String QUERY_TYPE = "query_type";
+    public static final String CONTENT_GENERATION_FIELD = "enable_Content_Generation";
     public static final String K_FIELD = "k";
     private final AbstractRetrieverTool queryTool;
     private String name = TYPE;
     private String description = DEFAULT_DESCRIPTION;
     private Client client;
     private String inferenceModelId;
+    private Boolean enableContentGeneration;
     private NamedXContentRegistry xContentRegistry;
-    private String index;
-    private String embeddingField;
-    private String[] sourceFields;
-    private String embeddingModelId;
     private String queryType;
-    private Integer docSize;
-    private Integer k;
     @Setter
     private Parser inputParser;
     @Setter
@@ -76,27 +75,14 @@ public class RAGTool extends AbstractRetrieverTool {
     public RAGTool(
         Client client,
         NamedXContentRegistry xContentRegistry,
-        String index,
-        String embeddingField,
-        String[] sourceFields,
-        Integer k,
-        Integer docSize,
-        String embeddingModelId,
         String inferenceModelId,
-        String queryType,
+        Boolean enableContentGeneration,
         AbstractRetrieverTool queryTool
     ) {
-        super(client, xContentRegistry, index, sourceFields, docSize);
         this.client = client;
         this.xContentRegistry = xContentRegistry;
-        this.index = index;
-        this.embeddingField = embeddingField;
-        this.sourceFields = sourceFields;
-        this.embeddingModelId = embeddingModelId;
-        this.docSize = docSize == null ? DEFAULT_DOC_SIZE : docSize;
-        this.k = k == null ? DEFAULT_K : k;
         this.inferenceModelId = inferenceModelId;
-        this.queryType = queryType;
+        this.enableContentGeneration = enableContentGeneration;
         this.queryTool = queryTool;
         outputParser = new Parser() {
             @Override
@@ -107,13 +93,6 @@ public class RAGTool extends AbstractRetrieverTool {
         };
     }
 
-    // getQueryBody is not used in RAGTool
-    @Override
-    protected String getQueryBody(String queryText) {
-        return queryText;
-    }
-
-    @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
         String input = null;
 
@@ -132,7 +111,9 @@ public class RAGTool extends AbstractRetrieverTool {
         String embeddingInput = input;
         ActionListener actionListener = ActionListener.<T>wrap(r -> {
             T queryToolOutput;
-
+            if (!this.enableContentGeneration) {
+                listener.onResponse(r);
+            }
             if (r.equals("Can not get any match from search result.")) {
                 queryToolOutput = (T) "";
             } else {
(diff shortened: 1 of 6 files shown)


mingshl:
known failure due to https://github.com/opensearch-project/flow-framework/issues/426

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -193,8 +187,30 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
-        vectorDBTool.run(Map.of(VectorDBTool.INPUT_FIELD, embeddingInput), actionListener);
 
+        Map<String, Object> params = new HashMap<>();
+        params.put(VectorDBTool.INDEX_FIELD, this.index);
+        params.put(VectorDBTool.EMBEDDING_FIELD, this.embeddingField);
+        params.put(VectorDBTool.SOURCE_FIELD, gson.toJson(this.sourceFields));
+        params.put(VectorDBTool.MODEL_ID_FIELD, this.embeddingModelId);
+        params.put(VectorDBTool.DOC_SIZE_FIELD, String.valueOf(this.docSize));
+
+        switch (this.queryType) {
```
zhichao-aws, line 198: How about moving the create NeuralSparseSearchTool/VectorDBTool logic to RAGTool.Factory.create? It is more logical consistent to create component during init process

zhichao-aws: And if there is malformed parameters, it makes more sense to throw it from "create" instead of "run"

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -193,8 +187,30 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
             log.error("Failed to search index.", e);
             listener.onFailure(e);
         });
-        vectorDBTool.run(Map.of(VectorDBTool.INPUT_FIELD, embeddingInput), actionListener);
 
+        Map<String, Object> params = new HashMap<>();
+        params.put(VectorDBTool.INDEX_FIELD, this.index);
```
zhichao-aws, line 192: How about using AbstractRetrieverTool.INDEX_FIELD? Same to below lines.

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -52,6 +52,7 @@ public class RAGTool extends AbstractRetrieverTool {
     public static final String EMBEDDING_MODEL_ID_FIELD = "embedding_model_id";
     public static final String EMBEDDING_FIELD = "embedding_field";
     public static final String OUTPUT_FIELD = "output_field";
+    public static final String QUERY_TYPE = "query_type";
```
zhichao-aws, line 55: Since we only use RAGTool in Olly, we may need a new parameter to control whether we directly return search results or call LLM to do summarize.

zhichao-aws (commented):
Thanks for your work! But I think it still need some changes.
Besides the comments for this PR, Since RAGTool doesn't call getQueryBody, and doesn't use the logic in AbstractRetrieverTool, I think we don't need to extend the AbstractRetrieverTool. Now the retrieval part has been done in RAGTool's component, i.e. VectorDBTool and NeuralSparseTool. It doesn't make sense to extend the AbstractRetrieverTool for RAGTool itself.

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: good call. It works the same, but I agree that the users input error can be checked in the earlier stage. Move the subtools(NeuralSparseSearchTool/VectorDBTool) initiation in the factory  create session.

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: No longer needed

mingshl:
> Thanks for your work! But I think it still need some changes. Besides the comments for this PR, Since RAGTool doesn't call getQueryBody, and doesn't use the logic in AbstractRetrieverTool, I think we don't need to extend the AbstractRetrieverTool. Now the retrieval part has been done in RAGTool's component, i.e. VectorDBTool and NeuralSparseTool. It doesn't make sense to extend the AbstractRetrieverTool for RAGTool itself.

I am more leaning towards keeping RAGTool extending from AbstractRetrieverTool, because it uses similar logic of creating a tool and run the parameters, it's also retrieving data through the sub tools(VectorDBTool and NeuralSparseTool). From the purpose of RAGTool, it retrieves data from subtool and use model for inference, so it should count towards a RetrieverTool as well.

mingshl:
fixing the JDK issue in the commit [Work around JDK 21.0.2 bug impacting scaling executors](https://github.com/opensearch-project/skills/pull/140/commits/b31617994b4f9c248eacf19db0efcfbbc76d60f7)

src/main/java/org/opensearch/agent/tools/RAGTool.java
zhichao-aws: Since we have moved subtools initiation in create session, we don't use the fields `index`, `embeddingField` ... in RAGTool.run now. We can remove these fields from RAGTool

src/main/java/org/opensearch/agent/tools/RAGTool.java
zhichao-aws: I mean we need a parameter to control whether to invoke the MLPredictionTaskAction, not how to interact with another model. Maybe users don't need further post-process for search results, but in current code we always do execute MLPredictionTaskAction after search. And we need a parameter decide whether RAGTool need to do this.

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -151,25 +149,25 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
             }
 
             Map<String, String> tmpParameters = new HashMap<>();
             tmpParameters.putAll(parameters);
 
-            if (vectorDBToolOutput instanceof List
-                && !((List) vectorDBToolOutput).isEmpty()
-                && ((List) vectorDBToolOutput).get(0) instanceof ModelTensors) {
-                ModelTensors tensors = (ModelTensors) ((List) vectorDBToolOutput).get(0);
+            if (queryToolOutput instanceof List
```
zhichao-aws, line 158: The `queryToolOutput`  is always String, right? Why we want to check if it's List or ModelTensor?

zhichao-aws:
> > Thanks for your work! But I think it still need some changes. Besides the comments for this PR, Since RAGTool doesn't call getQueryBody, and doesn't use the logic in AbstractRetrieverTool, I think we don't need to extend the AbstractRetrieverTool. Now the retrieval part has been done in RAGTool's component, i.e. VectorDBTool and NeuralSparseTool. It doesn't make sense to extend the AbstractRetrieverTool for RAGTool itself.
> 
> I am more leaning towards keeping RAGTool extending from AbstractRetrieverTool, because it uses similar logic of creating a tool and run the parameters, it's also retrieving data through the sub tools(VectorDBTool and NeuralSparseTool). From the purpose of RAGTool, it retrieves data from subtool and use model for inference, so it should count towards a RetrieverTool as well.

The core logic of AbstractRetrieverTool is `getQueryBody` and `run`, but we don't use `getQueryBody`, and override run. It doesn't make sense to extend an abstract class but doesn't use it's logic at all. If we just want to mark the retrieval purpose, maybe we can make them implements a new interface. Extending the AbstractRetrieverTool will confuse other contributors.

There is also another implementation if we do want to extend AbstractRetrieverTool. In getQueryBody, we generate `neural` or `neural_sparse` query, and in `run`, we call super.run() to do search.

zhichao-aws:
I also notice we still don't have integ test for RAGTool now. We can use `MockHttpServer` to mock an LLM endpoint to do integ test for RAGTool. You can check PPLToolIT for reference (https://github.com/opensearch-project/skills/blob/8074492891c67a72c1584d716d738d50979622a3/src/test/java/org/opensearch/integTest/PPLToolIT.java#L28). 

I don't think this is a blocker for this PR, but we need to add integ test in follow up PRs.

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: removed

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: added a parameter [enableContentGeneration to RAGTool](https://github.com/opensearch-project/skills/pull/140/commits/b12fd8ad2c9196f89d1978df7ba05dfdbf05b491)

src/main/java/org/opensearch/agent/tools/RAGTool.java
mingshl: removed

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -264,32 +233,20 @@ public RAGTool create(Map<String, Object> params) {
...
-                        .docSize(docSize)
                         .inferenceModelId(inferenceModelId)
-                        .queryType(queryType)
+                        .enableContentGeneration(enableContentGeneration)
                         .queryTool(neuralSparseSearchTool)
                         .build();
                 case "neural":
                     VectorDBTool.Factory.getInstance().init(client, xContentRegistry);
-                    params.put(VectorDBTool.MODEL_ID_FIELD, EMBEDDING_MODEL_ID_FIELD);
+                    params.put(VectorDBTool.MODEL_ID_FIELD, embeddingModelId);
```
zhichao-aws, line 242: We should also do this for NeuralSparseTool

zhichao-aws: We don't need to call `VectorDBTool.Factory.getInstance().init(client, xContentRegistry)` or `NeuralSparseSearchTool.Factory.getInstance().init(client, xContentRegistry)` here. This has been done during the initialization of skills plugin

mingshl: added the initiation to the test to simulate the initiations of tools in  [ToolPlugins.java](https://github.com/mingshl/skills/blob/5d7b96281e9c218416c9f85c60e6904bc074c1d7/src/main/java/org/opensearch/agent/ToolPlugin.java#L68) 

mapped the inference_model_id to model_id for NeuralSparseTool

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -186,33 +155,33 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
-    @Override
     public String getType() {
         return TYPE;
     }
 
     @Override
+    public String getVersion() {
+        return null;
+    }
+
```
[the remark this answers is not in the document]
mingshl: used `Factory implements Tool.Factory`

zhichao-aws (approved):
We can get PR this merged first, but may need to alter the prompt handling logic of RAGTool in the future based on e2e test result.

Need to fix the flaky test and add integration test for RAGTool in follow up PR.

PR #143, opened 2024-01-24
Include more return values for SearchAnomalyDetectorsTool and SearchMonitorsTool

ohltyler:
### Description
This PR includes a handful of improvements to the alerting and AD tools:
- includes more return fields to the anomaly detector and alerting monitor results
- improves UT to cover these added return values
- tunes descriptions of all 4 alerting & AD tools to list the individual return fields. This is necessary for an LLM to parse through the results better when a question may be asking about such fields (e.g., "When was detector 'A' last updated?", requiring to fetch detector 'A', then parse the 'lastUpdateTime' field).
 
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/agent/tools/SearchAlertsTool.java
```diff
@@ -33,7 +33,7 @@
...
...ault is 20), and startIndex which defines the index to start from (default is 0), and searchString which defines the search string to use for searching a specific alert (default is an empty String), and severityLevel which defines the severity level to filter for (default is ALL), and alertState which defines the alert state to filter for (default is ALL), and monitorId which defines the monitor ID to filter for, and alertIndex which defines the alert index to search from (default is null), and monitorIds which defines the list of monitor IDs to filter for, and workflowIds which defines the list of workflow IDs to filter for（default is null), and alertIds which defines the list of alert IDs to filter for (default is null). The tool returns 2 values: a list of alerts (each containining id, version, schema version, monitor ID, workflow ID, workflow name, monitor name, monitor version, monitor user, trigger ID, trigger name, finding IDs, related doc IDs, state, start time, end time, last notifcation time, acknowledged time, error message, error history, severity, action execution results, aggregation result bucket, execution ID, associated alert IDs), and the total number of alerts.";
```
lezzago, line 36: Did we see improvement where we defined what the alert object value contained? Do we have any information or metrics that showed doing this helps the tools as I see this is done for AD as well.

ohltyler: Not yet, that will be the next step. The main point here is that many questions are going to involve parsing responses to return fields within detectors/monitors/results/etc. (see example in description). For now, I'm starting out by at least exposing them, and tuning later as needed.

lezzago (approved):

Issue #136, opened 2024-01-18
Improve alerting tool IT


ohltyler:
Current alerting tool IT is rudimentary and does not test many cases. Add tests for both alerting tools to check:
- sample data in system indices is processed correctly
- filtering works correctly
- single piece of data is processed correctly
- multiple pieces of data is processed correctly

PR #151, opened 2024-01-26
Fix SearchMonitorsTool bugs; add corresponding ITs

ohltyler:
### Description
This PR fixes few bugs related to the SearchMonitorsTool:
- result parsing bug where monitor fields being `null` due to discrepancy on alerting plugin's transport vs. REST action for the search monitor API. TLDR is the REST action does post-processing by removing the nested `monitor` field, hence we need to do the same thing in the tool.
- search by `monitorID` bug due to classloading issue with the `getMonitor()` fn via common-utils. Simplified the tool to only use the search monitors transport action which improves maintainability as well

Also fixes a bug in SearchAnomalyDetectorsTool where the mapping should be `detector_type` instead of `type`.
 
Also adds several IT cases to cover the bugs. Completes part of #136 
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java
+++ b/src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java
@@ -20,7 +20,6 @@ import org.opensearch.client.Client;
 import org.opensearch.client.node.NodeClient;
 import org.opensearch.commons.alerting.AlertingPluginInterface;
 import org.opensearch.commons.alerting.action.SearchMonitorRequest;
-import org.opensearch.commons.alerting.model.Monitor;
 import org.opensearch.commons.alerting.model.ScheduledJob;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.index.query.BoolQueryBuilder;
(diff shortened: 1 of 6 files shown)


ohltyler:
Known failures from neural sparse tool. Added IT and all UT pass on CI and locally.
```
30 tests completed, 2 failed
Tests with failures:
 - org.opensearch.integTest.NeuralSparseSearchToolIT.testNeuralSparseSearchToolInFlowAgent_withIllegalEmbeddingField_thenThrowException
 - org.opensearch.integTest.NeuralSparseSearchToolIT.testNeuralSparseSearchToolInFlowAgent_withIllegalIndexField_thenThrowException
```

src/test/java/org/opensearch/integTest/SearchMonitorsToolIT.java
```diff
@@ -54,5 +60,86 @@ public void testSearchMonitorsToolInFlowAgent_withNoSystemIndex() {
         assertEquals("Monitors=[]TotalMonitors=0", result);
     }
 
-    // TODO: Add IT to test against sample monitor data
+    @SneakyThrows
+    public void testSearchMonitorsToolInFlowAgent_searchById() {
+        deleteSystemIndices();
```
lezzago, line 65: Should we add this in a @BeforeEach?

src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java
```diff
@@ -102,89 +98,73 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
...
+                "monitor.triggers",
+                new ExistsQueryBuilder("monitor.triggers"),
+                ScoreMode.None
+            );
+
+            BoolQueryBuilder triggerQuery = new BoolQueryBuilder();
+            if (hasTriggers) {
+                triggerQuery.must(nestedTriggerQuery);
+            } else {
+                triggerQuery.mustNot(nestedTriggerQuery);
```
lezzago, line 125: If we are adding these filters to find the monitors, I think it would be beneficial to also search by monitor type as well.

ohltyler: This PR isn't adding any filters, it's just shown as updated due to tabbing since the overall if/else was removed. Prefer to leave the tool with the current set of params and add/remove/tune later on as more testing is done and we capture the breadth of questions and functional responses.

src/test/java/org/opensearch/integTest/SearchMonitorsToolIT.java
ohltyler: Sure, added

src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java
ohltyler: Especially for initial release, we need to be weary about how many knobs we allow the LLM to tune. Ideally we cover as many questions as possible in a confident/consistent/accurate way, and slowly adjust or expand over time after we have a good gauge of performance.

ohltyler:
^ latest commit fixes a bug found in integ tests where the mapping is `detector_type` instead of `type`, causing `null` values set for it. This commit updates that.

src/main/java/org/opensearch/agent/tools/SearchMonitorsTool.java
lezzago: Fair enough, we can set this as a further enhancement, down the road.

lezzago (approved):

PR #184, opened 2024-02-05
(PPLTool) update claude and openai prompts

joshuali925:
### Description
The existing prompt for claude doesn't work, since it's using `%s` as place holders. Additionally there were too many escapes.

This PR fixes it and adds OpenAI prompt, which is the same as claude prompt except PPL in the examples are surrounded by `<ppl>` `</ppl>` tags. This prompt might get updated to tune performance later.
 
### Issues Resolved
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -130,7 +130,8 @@ public class PPLTool implements Tool {
 
     public enum PPLModelType {
         CLAUDE,
-        FINETUNE;
+        FINETUNE,
+        OPENAI;
```
joshuali925, line 135: make the model type more specific and include model version

xinyual (approved):

PR #187, opened 2024-02-06
feat: use original input for RAGTool / VectorDBTool

SuZhou-Joe:
### Description
In conversational agent, LLM is transforming user's input to other input like JSON / DSL query etc. So for VectorDBTool / RAGTool, we need to keep original input.
 
### Issues Resolved
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/RAGTool.java
+++ b/src/main/java/org/opensearch/agent/tools/RAGTool.java
@@ -71,6 +71,11 @@ public class RAGTool implements Tool {
     @Setter
     private Parser outputParser;
 
+    @Override
+    public boolean useOriginalInput() {
+        return true;
+    }
+
     @Builder
     public RAGTool(
         Client client,
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -50,6 +50,11 @@ public abstract class AbstractRetrieverTool implements Tool {
     protected Integer docSize;
     protected String version;
 
+    @Override
+    public boolean useOriginalInput() {
+        return true;
+    }
+
     protected AbstractRetrieverTool(
         Client client,
         NamedXContentRegistry xContentRegistry,
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -50,11 +50,6 @@ public abstract class AbstractRetrieverTool implements Tool {
     protected Integer docSize;
     protected String version;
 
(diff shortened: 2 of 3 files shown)


src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -71,6 +71,11 @@ public class RAGTool implements Tool {
     @Setter
     private Parser outputParser;
 
+    @Override
```
zane-neo, line 74: This is a new method defined in Tool's interface?

SuZhou-Joe: Nope, it's an existing method but none of the tools override it. For RAGTool and VectorDBTool, they need original input(plain text) instead of a LLM-transformed input.

zane-neo: Can you add some simple comments to describe why original input is needed for this two tools?

xinyual: Does Agent framework support this now?

src/main/java/org/opensearch/agent/tools/RAGTool.java
```diff
@@ -71,6 +71,11 @@ public class RAGTool implements Tool {
...
+    public boolean useOriginalInput() {
```
mingshl, line 75: Thanks @SuZhou-Joe to help flipping on the flag following the fix in common https://github.com/opensearch-project/ml-commons/pull/2022 using original input for RAGTool/VectorDBTool/NeuralSparseSearchTool. IMO, the neuralSparseSearchTool also need to take original input question instead of llm generated questions. @zhichao-aws would you agree on this for NeuralSparseSearchTool?

In this case, since vectorDB and NeuralSparseSearchTool are both extending from AbstractRetrieverTool, we can modify the AbstractRetrieverTool to flip on UseOriginalInput to true. Then both VectorDB and NeuralSparseSearchTool are using original input. 

RAGTool also need to keep this useOrginalInput flag because it's not extending from AbstractRetrieverTool now.

SuZhou-Joe:
Hi team, we are trying to do a JSON serialization when finding that the input question is a JSON syntax @zhichao-aws . The new solution has some pros:
- It can still leverage the LLM's output to increase search accuracy.
- It will address the issue that if we keep the original input and input happens to have" inside, the `buildQuery` method will throw error as it break the structure of a standard JSON.

zhichao-aws:
The root cause is we format a query string (json string) in search tools. The query text is expected to be a plain text. If the query text itself is json string, it will throw exceptions to parse the formated string.  Fix this in #203

zhichao-aws:
#203 is merged now. Let's close this PR. We can re-open it on-demand.

PR #206, opened 2024-02-07
add truncate

xinyual:
### Description
Sometimes PPL will return too much information, we need a truncation to limit the return length of PPL tool.
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.agent.tools;
 
+import static org.apache.commons.lang3.math.NumberUtils.min;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
@@ -94,6 +96,8 @@ public class PPLTool implements Tool {
 
     private String previousToolKey;
 
+    private int truncate;
+
     private static Gson gson = new Gson();
 
     private static Map<String, String> DEFAULT_PROMPT_DICT;
@@ -147,7 +151,15 @@ public class PPLTool implements Tool {
 
     }
 
-    public PPLTool(Client client, String modelId, String contextPrompt, String pplModelType, String previousToolKey, boolean execute) {
+    public PPLTool(
+        Client client,
+        String modelId,
+        String contextPrompt,
+        String pplModelType,
+        String previousToolKey,
+        int truncate,
+        boolean execute
+    ) {
         this.client = client;
         this.modelId = modelId;
         this.pplModelType = PPLModelType.from(pplModelType);
@@ -157,6 +169,7 @@ public class PPLTool implements Tool {
             this.contextPrompt = contextPrompt;
         }
         this.previousToolKey = previousToolKey;
+        this.truncate = truncate;
         this.execute = execute;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -488,6 +500,14 @@ private String parseOutput(String llmOutput, String indexName) {
         }
         ppl = ppl.replace("`", "");
         ppl = ppl.replaceAll("\\bSPAN\\(", "span(");
+        if (this.head > 0) {
+            String[] lists = llmOutput.split("\\|");
```
zane-neo, line 504: Can we simply use contains method to check if llmOutput has head, if true do not modify it. This is more efficient than splitting and then check.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -488,6 +500,14 @@ private String parseOutput(String llmOutput, String indexName) {
...
+            String lastCommand = lists[lists.length - 1].strip();
```
zane-neo, line 505: Is it possible the llmOutput ends with `|`, will this case cause NPE here?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -308,6 +319,7 @@ public PPLTool create(Map<String, Object> map) {
                 (String) map.getOrDefault("prompt", ""),
                 (String) map.getOrDefault("model_type", ""),
                 (String) map.getOrDefault("previous_tool_name", ""),
+                Integer.valueOf((String) map.getOrDefault("head", "-1")),
```
zane-neo, line 322: Can we make this a non-optional field or give it a default value, when it's -1 and the response is long, this is still possible to exceed the max tokens to llm.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: but head can be the field name. Split can be the safe way.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: Will add a check on that.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: I prefer to let customer to define this. The head number should be related to the data inside index. If the data like log, is very long, the head number should be small. If we want to set a default number for that, it should be a larger one without breaking some correct answer. So I prefer a large number like 200 or just leave as optional one.

src/main/java/org/opensearch/agent/tools/PPLTool.java
zane-neo: Will the `lists[lists.length - 1]` return null?

xinyual: Even a empty string with `.split("\\|")` will return a one size list [""], so I think it won't be null.

xinyual: The only possible it will fail is the output only contain | like "|||", but it will raise the IllegalArgumentException before we add head: https://github.com/opensearch-project/skills/blob/7c2ccf8aaad6552fb04cf17ab41643ffc736266a/src/main/java/org/opensearch/agent/tools/PPLTool.java#L486

zane-neo:
CI is failing, please take a look.

xinyual:
> CI is failing, please take a look.

Already fixed

zane-neo (approved):

PR #233, opened 2024-02-19
Fix SearchAnomalyDetectorsTool indices param bug; add more IT

ohltyler:
### Description
This PR fixes an edge case bug of the `indices` param failing for the `SearchAnomalyDetectorsTool` if the input contains hyphens (e.g., `sample-ecommerce-index`). By default this breaks term queries as this splits this up into multiple terms by the standard query analyzer. This fixes that by searching on the indices keyword instead to remove the complexity and potential failures.

Adds IT to cover this case plus adds more coverage of other params on matching / non-matching scenarios.

Additionally, cleans up the create agent request bodies to omit overriding the default tool descriptions which have much more detail.
 
### Issues Resolved
 
### Check List


joshpalis (approved):

owaiskazi19 (approved):
LGTM

PR #235, opened 2024-02-21
Fix detector state params in SearchAnomalyDetectorsTool

ohltyler:
### Description

This PR fixes two issues related to fetching and filtering detector state with the `SearchAnomalyDetectorsTool`.

#### `NamedWriteableRegistry` error
This fix coincides with https://github.com/opensearch-project/anomaly-detection/pull/1164 - see there for more details on the original issue.

This change injects the `NamedWriteableRegistry` to the AD-related tools, in order to construct the AD node client with its updated constructor params that now requires a `NamedWriteableRegistry`. At a high level, this is used in the node client for being able to successfully serialize/deserialize the `GetAnomalyDetectorResponse` used when fetching detector profiles, when the `SearchAnomalyDetectorsTool` is invoked where any state-related param is passed down - e.g., `running`/`failed`.

#### Detectors filtering bug
There is an issue (uncovered after the above fix was made causing the tests to fail) of detectors with mismatching states not being filtered out. The root cause was that the profile detector transport action response does not include the detector ID by default in the `anomalyDetector` field, hence it was always `null` and nothing was removed from the hits map. This changes that by using the detector name as the key in the hits map, which is returned by default.

#### Miscellaneous
- Removes `disabled` as a parameter. This caused LLM confusion, since a question like "how many stopped detectors do i have?" could lead to setting the parameters as `"disabled": "true"`, or `"running": "false"`. This simplifies that by only having one knob to turn instead of two. Additionally, this changes simplifies the filtering logic and the matrix combination of values shrink.
- Other filtering logic cleanup
- Adds integration tests to test out the detector state params and enforces filtering works as expected
- Cleans up existing integration tests to handle individual-test-failures more cleanly and less error logs when running the tests
- Cleans up existing unit tests
- Minor cleanup to search AD results ITs to handle various flakiness

The overall impact of these changes is that detector-state-related questions passed to an agent provisioned with the `SearchAnomalyDetectorsTool` now works as expected. 
 
### Issues Resolved
 
### Check List
  - [x] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/ToolPlugin.java
+++ b/src/main/java/org/opensearch/agent/ToolPlugin.java
@@ -69,8 +69,8 @@ public class ToolPlugin extends Plugin implements MLCommonsExtension {
         SearchIndexTool.Factory.getInstance().init(client, xContentRegistry);
         RAGTool.Factory.getInstance().init(client, xContentRegistry);
         SearchAlertsTool.Factory.getInstance().init(client);
-        SearchAnomalyDetectorsTool.Factory.getInstance().init(client);
-        SearchAnomalyResultsTool.Factory.getInstance().init(client);
+        SearchAnomalyDetectorsTool.Factory.getInstance().init(client, namedWriteableRegistry);
+        SearchAnomalyResultsTool.Factory.getInstance().init(client, namedWriteableRegistry);
         SearchMonitorsTool.Factory.getInstance().init(client);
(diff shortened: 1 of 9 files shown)


ohltyler:
Leaving as draft until the related AD PR is merged so the CI can pick up the updated sonatype dependency.

joshpalis (approved):

amitgalitz (approved):

PR #259, opened 2024-03-14
add parameter validate for PPL tool

xinyual:
### Description
add parameter validate for PPL tool: 
1. Model id should not be null or blank
2. Head should be integer
3. Execute should be false or true
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -18,6 +18,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.StringJoiner;
 import java.util.regex.Matcher;
@@ -362,7 +363,7 @@ public class PPLTool implements Tool {
         if (StringUtils.isBlank((String) map.get("model_id"))) {
             throw new IllegalArgumentException("PPL tool needs non blank model id.");
         }
-        if (map.containsKey("execute") && map.get("execute") != null) {
+        if (map.containsKey("execute") && Objects.nonNull(map.get("execute"))) {
             String execute = map.get("execute").toString().toLowerCase(Locale.ROOT);
             if (!execute.equals("true") && !execute.equals("false")) {
                 throw new IllegalArgumentException("PPL tool parameter execute must be false or true");
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -23,7 +23,6 @@ import java.util.StringJoiner;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.math.NumberUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -356,6 +358,32 @@ private GetMappingsRequest buildGetMappingRequest(String indexName) {
         return getMappingsRequest;
     }
 
+    private static void validatePPLToolParameters(Map<String, Object> map) {
```
zane-neo, line 361: Looks like this if...else can be merged to `StringUtils.isEmpty(map.get("model_id"))`?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -356,6 +358,32 @@ private GetMappingsRequest buildGetMappingRequest(String indexName) {
...
+            throw new IllegalArgumentException("PPL tool needs model id.");
+        } else {
+            String modelId = (String) map.get("model_id");
+            if (StringUtils.isBlank(modelId)) {
+                throw new IllegalArgumentException("PPL tool needs non blank model id.");
+            }
+        }
+        if (map.containsKey("execute")) {
+            String execute = map.get("execute").toString().toLowerCase(Locale.ROOT);
+            if (!(execute.equals("true")) && !(execute.equals("false"))) {
```
zane-neo, line 372: redundant `()`, can be `!execute.equals("true")`

zane-neo, line 372: redundant `()`, can be `!execute.equals("true")`

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -356,6 +358,32 @@ private GetMappingsRequest buildGetMappingRequest(String indexName) {
...
+                throw new IllegalArgumentException("PPL tool parameter execute must be false or true");
+            }
+
+        }
+        if (map.containsKey("head")) {
+            String head = map.get("head").toString();
+            try {
+                int headInt = NumberUtils.createInteger(head);
```
zane-neo, line 380: Do we need to check if the head is positive?

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -356,6 +358,27 @@ private GetMappingsRequest buildGetMappingRequest(String indexName) {
         return getMappingsRequest;
     }
 
+    private static void validatePPLToolParameters(Map<String, Object> map) {
+        if (StringUtils.isBlank((String) map.get("model_id"))) {
+            throw new IllegalArgumentException("PPL tool needs non blank model id.");
+        }
+        if (map.containsKey("execute")) {
+            String execute = map.get("execute").toString().toLowerCase(Locale.ROOT);
```
zane-neo, line 366: Is it possible the `map.get("execute")` is null? if yes, it could produce NPE here.

src/main/java/org/opensearch/agent/tools/PPLTool.java
xinyual: I change the logic to prevent it. Please check it. From my perspective, these parameters are from register body and couldn't be null. But anyway, it's better to have it.

zane-neo (approved):

PR #344, opened 2024-07-03
Add cluster setting to control ppl execution

zane-neo:
### Description
Add cluster setting to control ppl execution
 
### Issues Resolved
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -226,7 +226,6 @@ public class PPLTool implements Tool {
                         return;
                     }
                     String ppl = parseOutput(dataAsMap.get("response"), indexName);
-                    System.out.println("###################### current ppl execution enabled value is: " + clusterSettingHelper.getClusterSettings(SkillSettings.PPL_EXECUTION_ENABLED) + ", execute value is: " + execute + " ######################");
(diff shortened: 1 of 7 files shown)


xinyual:
We already have a Tool level execution controller, why we need a cluster level one? Is this for OASIS?

zane-neo:
> We already have a Tool level execution controller, why we need a cluster level one? Is this for OASIS?

Yes, in serverless currently the ppl execution will be disabled by default.

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -227,7 +226,7 @@ public <T> void run(Map<String, String> parameters, ActionListener<T> listener)
                         return;
                     }
                     String ppl = parseOutput(dataAsMap.get("response"), indexName);
-                    if (!this.execute) {
+                    if (!clusterSettingHelper.getClusterSettings(SkillSettings.PPL_EXECUTION_ENABLED) || !this.execute) {
                         Map<String, String> ret = ImmutableMap.of("ppl", ppl);
```
xinyual, line 230: Can we add a log here when cluster setting and tool setting are conflict? For example, "To execute PPL, please set lugins.skills.ppl_execution_enabled to true". It may be confused if customer set tool setting to true but it is not executed.

zane-neo: Sure, will add debug log here.

xinyual (approved):

PR #348, opened 2024-07-10
Add CreateAnomalyDetectorTool

gaobinlong:
### Description

Implement a new tool named `CreateAnomalyDetectorTool` which can be used to help creating anomaly detector, the only one input parameter is the index name, this tool will get the mappings of the index in flight and let LLM give the suggested category field, aggregation field and correspond aggregation method which are required for the create anomaly detector API, the output of this tool is like:
 ```
{
      "index": "opensearch_dashboards_sample_data_ecommerce",
      "categoryField": "geoip.country_iso_code",
      "aggregationField": "total_quantity,total_unique_products,taxful_total_price",
      "aggregationMethod": "sum,count,sum",
      "dateFields": "customer_birth_date,order_date,products.created_on"
 }
```

### Related Issues

https://github.com/opensearch-project/skills/issues/337

### Check List

### CODE REVIEW

**Comment by zane-neo** [comment]
@gaobinlong please fix the conflict

**Comment by gaobinlong** [comment]
> @gaobinlong please fix the conflict

Done it, thanks!

File: src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java [hunk 1]
```diff
@@ -0,0 +1,413 @@
...
+                MLInput.builder().algorithm(FunctionName.REMOTE).inputDataset(inputDataSet).build(),
+                null
+            );
+
+            client.execute(MLPredictionTaskAction.INSTANCE, request, ActionListener.wrap(mlTaskResponse -> {
+                ModelTensorOutput modelTensorOutput = (ModelTensorOutput) mlTaskResponse.getOutput();
+                ModelTensors modelTensors = modelTensorOutput.getMlModelOutputs().get(0);
+                ModelTensor modelTensor = modelTensors.getMlModelTensors().get(0);
+                Map<String, String> dataAsMap = (Map<String, String>) modelTensor.getDataAsMap();
+                String finalResponse = dataAsMap.get("response");
```
zane-neo, line 211: This can be optimized to use Optional to extract the response to avoid NPE.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -0,0 +1,413 @@
...
+     * @param indexName the index name
+     * @return the prompt about creating anomaly detector
+     */
+    private String constructPrompt(final Map<String, String> fieldsToType, final String indexName) {
+        StringJoiner tableInfoJoiner = new StringJoiner("\n");
+        for (Map.Entry<String, String> entry : fieldsToType.entrySet()) {
+            tableInfoJoiner.add("- " + entry.getKey() + ": " + entry.getValue());
+        }
+
+        return "Human:\" turn\": Here is an example of the create anomaly detector API: POST _plugins/_anomaly_detection/detectors "
```
zane-neo, line 318: Can we move this prompt to a file for easier reading and reduced maintenance effort?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -0,0 +1,413 @@
...
+         * @return the instance of this tool
+         */
+        @Override
+        public CreateAnomalyDetectorTool create(Map<String, Object> map) {
+            Object obj = map.get("model_id");
+            if (Objects.isNull(obj)) {
+                throw new IllegalArgumentException("model_id cannot be null.");
+            }
+            String modelId = (String) obj;
+            if (modelId.isEmpty()) {
```
zane-neo, line 392: This can potentially throw NPE.

xinyual:
This tool will not actually create AD but only return suggested parameters, right?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
xinyual: You can put this into src/resources

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: You mean `dataAsMap` can be null?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: Sure, I'll do that.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: `obj` was checked in line 388.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -0,0 +1,413 @@
...
+            result.putAll(chatParameters);
+        } catch (Exception e) {
+            // input is a string
+            String indexName = parameters.getOrDefault("input", "");
+            if (!indexName.isEmpty()) {
+                result.put("index", indexName);
+            }
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
gaobinlong: If `parameters` is an immutable map, then modifying it will throw `UnsupportedException`.

gaobinlong:
> This tool will not actually create AD but only return suggested parameters, right?

Yes. @zane-neo @xinyual  I've made some changes following your suggestions, please help to take a look, thank you!

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
zane-neo: Yes, this could be null.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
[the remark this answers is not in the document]
zane-neo: make sense.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -0,0 +1,452 @@
...
+                        .onFailure(
+                            new IllegalStateException(
+                                "The inference result from remote endpoint is not valid, cannot extract the key information from the result."
+                            )
+                        );
+                    return;
+                }
+
+                // remove double quotes or whitespace if exists
+                String categoryField = matcher.group(1).replaceAll("\"", "").strip();
```
zane-neo, line 268: Is it possible the `matcher.group(x)` returns null?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -0,0 +1,452 @@
...
+                    );
+                listener.onResponse((T) AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> gson.toJson(result)));
+            }, e -> {
+                log.error("fail to predict model: " + e);
+                listener.onFailure(e);
+            }));
+        }, e -> {
+            log.error("fail to get mapping: " + e);
+            String errorMessage = e.getMessage();
+            if (errorMessage.contains("no such index")) {
```
zane-neo, line 293: It's better to check exception type instead of error message, since error message isn't a contract and might change in future.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: If the string matches the regex, then `matcher.group(x) ` couldn't be null.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: Changed that, please help to take a look, thank you!

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
gaobinlong: Add a check before `dataAsMap.get()`.

zane-neo (approved):

src/main/java/org/opensearch/agent/tools/utils/ToolHelper.java
```diff
@@ -0,0 +1,42 @@
...
+                Map<String, Object> vMap = (Map<String, Object>) v;
+                if (vMap.containsKey("type")) {
+                    if (!((vMap.getOrDefault("type", "")).equals("alias"))) {
+                        fieldsToType.put(prefix + n, (String) vMap.get("type"));
+                    }
+                }
+                if (vMap.containsKey("properties")) {
+                    extractFieldNamesTypes((Map<String, Object>) vMap.get("properties"), fieldsToType, prefix + n);
+                }
+                if (vMap.containsKey("fields")) {
```
gaobinlong, line 36: This method `extractFieldNamesTypes` leverages most of the code of the original method in the PPLTool class, but I made some change because the following cases should be taken into account:
```
"a": {
      "type": "object",
      "properties": {
        "b": {
          "type":"keyword"
        }
      }
    }
```
and 
```
"c": {
      "type": "text",
      "fields": {
        "d": {
          "type":"keyword"
        }
      }
    }
```
, @zane-neo and @xinyual, please help to check if this change has a bad impact on the PPLTool,  thank you!

zane-neo: It seems after this change, the recursion will go one layer deeper when `type` is `text`, which results in changing the result from `text` to `keyword`, @xinyual please check from your end if this have impact on PPLTool.

Issue #338, opened 2024-06-18
[FEATURE] Create alerting monitor tool


ylwu-amzn:
We have get alerting monitor tool now. We can create some tool to create alerting monitor.

dblock:
[Catch All Triage, attendees [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/lukas-vlcek/), [4](https://github.com/rafaelma), [5](https://github.com/krisfreedain), [6](https://github.com/gargrajat), [7](https://github.com/andrross)]

qianheng-aws:
Hi, I'm working on implementing this tool, which aims to help create alert with natural language. The tool has 3 input parameters:
  1. question: user's question about creating a new alert
  2. indices: The input indices of the monitor.
  3. chat_history: optional, the chat history of chatbot

This tool is able to request LLM by itself and has its own tool prompt, and its can also get necessary information by calling OS API.

When this tool is selected by chat agent or invoked by flow agent, it will be executed as following steps:
 1. get the mapping info of the input indices
 2. construct its final tool prompt by substitute the placeholders(e.g. question, indices, chat_history, mapping_info) in its raw prompt
 3. request LLM with the final prompt and got response
 4. extract the alert configuration from the LLM response as its output.
 
 The tool output format may be like:
```
Tool output:

{
    "name": "Error 500 Response Alert",
    "search": {
        "indices": ["opensearch_dashboards_sample_data_logs"],
        "timeField": "timestamp",
        "bucketValue": 60,
        "bucketUnitOfTime": "m",
        "filters": [
            {
                "fieldName": [
                    {
                        "label": "response",
                        "type": "text"
                    }
                ],
                "fieldValue": "500",
                "operator": "is"
            }
        ],
        "aggregations": [
            {
                "aggregationType": "count",
                "fieldName": "bytes"
            }
        ]
    },
    "triggers": [
        {
            "name": "Error 500 Response Count Above 1",
            "severity": 1,
            "thresholdValue": 1,
            "thresholdEnum": "ABOVE"
        }
    ]
}
```
There is a demo video about creating an alert by LLM: https://broadcast.amazon.com/videos/1159580

[...]

PR #349, opened 2024-07-12
add CreateAlertTool

qianheng-aws:
### Description
Implementing CreateAlertTool, which aims to help create alert with natural language. The tool has 3 input parameters:

- question: user's question about creating a new alert
- indices: The input indices of the monitor.
- chat_history: optional, the chat history of chatbot. This is helpful when creating alert as a follow-up question in conversation.

This tool is able to request LLM by itself and has its own tool prompt, and it can also get some necessary information by calling OS API.

When this tool is selected by chat agent or invoked by flow agent, it will be executed as following steps:

- get the mapping info of the input indices
- construct its final tool prompt by substitute the placeholders(e.g. question, indices, chat_history, mapping_info) in its raw prompt
- request LLM with the final prompt and got response
- extract the alert configuration from the LLM response as its output.

The tool output format may be like:
```
Tool output:
{
    "name": "Error 500 Response Alert",
    "search": {
        "indices": ["opensearch_dashboards_sample_data_logs"],
        "timeField": "timestamp",
        "bucketValue": 60,
        "bucketUnitOfTime": "m",
        "filters": [
            {
                "fieldName": [
                    {
                        "label": "response",
                        "type": "text"
                    }
                ],
                "fieldValue": "500",
                "operator": "is"
            }
        ],
        "aggregations": [
            {
                "aggregationType": "count",
                "fieldName": "bytes"
            }
        ]
    },
    "triggers": [
        {
            "name": "Error 500 Response Count Above 1",
            "severity": 1,
            "thresholdValue": 1,
            "thresholdEnum": "ABOVE"
        }
    ]
}
```

**Note**: it relies on ChatAgentRunner to pass chat_history to this tool, so it needs to add a new method in Tool interface and override it, as explained in this issue #[2645](https://github.com/opensearch-project/ml-commons/issues/2645)

There is a demo video about creating an alert by LLM: https://broadcast.amazon.com/videos/1159580

### Related Issues
Resolves #[[338](https://github.com/opensearch-project/skills/issues/338)]

Has dependency on this issue in ml-commons repo: #[2645](https://github.com/opensearch-project/ml-commons/issues/2645)

### Check List


src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    private String getIndexMappingInfo(Map<String, String> parameters)
+        throws InterruptedException, ExecutionException {
+        if (!parameters.containsKey("indices") || parameters.get("indices").isEmpty()) {
+            throw new IllegalArgumentException("No indices in the input parameter. Ask user to "
+                + "provide index as your final answer directly without using any other tools");
+        }
+        String rawIndex = parameters.getOrDefault("indices", "");
```
xinyual, line 206: Please check whether the index doesn't start with '.' since it is system index we cannot query like here https://github.com/opensearch-project/skills/blob/14d9ef25a8f63b93d7fcbe5e6a725941b342ae4a/src/main/java/org/opensearch/agent/tools/PPLTool.java#L191

zane-neo:
@qianheng-aws please run `./gradlew spotlessApply` to format code

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,282 @@
...
+    private String name = TYPE;
+    @Getter
+    @Setter
+    private String description = DEFAULT_DESCRIPTION;
+
+    private final Client client;
+    private final String modelId;
+    private final String TOOL_PROMPT_TEMPLATE;
+
+    private static final Gson gson = new Gson();
```
zane-neo, line 71: Please use `org.opensearch.ml.common.utils.StringUtils.gson` to avoid creating new instance.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,282 @@
...
+                listener.onFailure(e);
+            }));
+        } catch (Exception e) {
+            log.error("Failed to call CreateAlertTool", e);
+            listener.onFailure(e);
+        }
+    }
+
+    private String getIndexMappingInfo(Map<String, String> parameters) throws InterruptedException, ExecutionException {
+        if (!parameters.containsKey("indices") || parameters.get("indices").isEmpty()) {
```
zane-neo, line 169: When the parameters contains the key but value is null, `parameters.get("indices").isEmpty()` can throw NPE.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,282 @@
...
+            );
+        }
+        String rawIndex = parameters.getOrDefault("indices", "");
+        List<String> indexList;
+        try {
+            indexList = gson.fromJson(rawIndex, new TypeToken<List<String>>() {
+            }.getType());
+        } catch (Exception e) {
+            // LLM sometimes returns the indices as a string instead of a json list, although we require that in the tool description.
+            indexList = Collections.singletonList(rawIndex);
```
zane-neo, line 182: Is rawIndex a comma separated string? What data structure are expected to stored in the `indexLIst`: ["hello, world"] or ["hello"], ["world"]?

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,282 @@
...
+                        rawIndex
+                    )
+            );
+        }
+        final String[] indices = indexList.toArray(Strings.EMPTY_ARRAY);
+        final GetIndexRequest getIndexRequest = new GetIndexRequest()
+            .indices(indices)
+            .indicesOptions(IndicesOptions.strictExpand())
+            .local(Boolean.parseBoolean(parameters.getOrDefault("local", "true")))
+            .clusterManagerNodeTimeout(DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT);
```
zane-neo, line 204: It's not preferred to blocking wait in code, unless you can confirm it's not running in event loop thread, also, what's the value of DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT?

src/main/java/org/opensearch/agent/tools/utils/ToolHelper.java
```diff
@@ -5,9 +5,38 @@
...
+import java.util.HashMap;
 import java.util.Map;
 
+import com.google.gson.Gson;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
 public class ToolHelper {
+    private static final Gson gson = new Gson();
```
zane-neo, line 20: same above

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
qianheng-aws: We expects rawIndex to be json string like `[index_name1]` mostly or `[index_name1, index_name2]` if there are 2 indices. Just as I tell LLM in the tool description `"1. indices: The input indices of the monitor, should be a list of string in json format`. So we can parse it to indexList via gson directly.

But there is one time LLM returns indices without `[]` which is only `index_name1`, so I catch the parse failing exception and use rawIndex directly here.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
qianheng-aws: Sorry, what do you mean of blocking wait? I just imitate what IndexMapping do here.

https://github.com/opensearch-project/ml-commons/blob/5ad26df7dd23a6ed37a13d80548779579521b13f/ml-algorithms/src/main/java/org/opensearch/ml/engine/tools/IndexMappingTool.java#L139

DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT is 30s.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
qianheng-aws: There is a UT in CreateAlertToolTests::testTool to cover this corner case

zane-neo: Is it possible that LLM returns "index1, index2" ? I suggest do a string split with comma here to cover this case as well.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
zane-neo: Line 205: client.admin().indices().getIndex(getIndexRequest)`.get()`, this get blocks the thread until result returns, it's preferred to use `getIndex(getIndexRequest, actionListener)` method instead of `getIndex(getIndexRequest)`.

qianheng-aws: I see. But it does need to retrieve index mapping info before requesting LLM. if using `getIndex(getIndexRequest, actionListener)`, then all logic after that should be put in the listener and wait for response from getIndex, that makes not difference with current implementation.

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
qianheng-aws: I haven't see this corner case so far. Will cover this case as well

zane-neo (approved):

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -0,0 +1,300 @@
...
+    private final String TOOL_PROMPT_TEMPLATE;
+
+    private static final String MODEL_ID = "model_id";
+    private static final String PROMPT_FILE_PATH = "CreateAlertDefaultPrompt.json";
+    private static final String DEFAULT_QUESTION = "Create an alert as your recommendation based on the context";
+
+    public CreateAlertTool(Client client, String modelId, String modelType) {
+        this.client = client;
+        this.modelId = modelId;
+        Map<String, String> promptDict = ToolHelper.loadDefaultPromptDictFromFile(this.getClass(), PROMPT_FILE_PATH);
```
xinyual, line 75: Can we make it static? Otherwise each time the agent executing, it will initialize the instance and load the json one more time.

xinyual (approved):

PR #413, opened 2024-09-26
Add LogPatternTool

qianheng-aws:
### Description
Add LogPatternTool to support generate log patterns based on an input `DSL`.

#### Background of log pattern 
Log pattern is a feature in Observability. It aims to find the common feature among retrieved logs.

The implementation of the log pattern feature in Observability includes 3 steps:

1. execute a PPL to retrieve the first log
2. set default pattern field, which is implemented in front-end. The default pattern field is found by scanning over the values of the first sample log. The field with the longest length of value will be the default pattern field. ([Code Ref](https://github.com/opensearch-project/dashboards-observability/blob/279463ad0c8e00780bf57f85e6d2bf30e8c7c93d/public/components/event_analytics/hooks/use_fetch_patterns.ts#L137C2-L164C6); For alert analysis, it seems meaningless to find pattern on the field with the longest length of value.)
3. execute a PPL to retrieve data and fill in log pattern table. Assuming field message is the pattern field, an example of such an PPL is:
    `source=opensearch_dashboards_sample_data_flights | where timestamp >= ’[2024-09-09 16](tel:2024090916):16:30.321000' and timestamp <= '2024-09-10 07:16:30.322000' | patterns Dest | stats count(), take(Dest, 1) by patterns_field` 
    By default, the operator pattern in PPL will generate a new field named pattern_field which will ignore characters conatined by [a-zA-Z\d] in the original value([Code Ref](https://github.com/opensearch-project/sql/blob/4303a2ab755d53903094dd94a5100572677a27a1/core/src/main/java/org/opensearch/sql/expression/parse/PatternsExpression.java#L27); For alert analysis, it also seems less helpful to analysis on such pattern). 

However, if customers use `DSL` to retrieve logs, it doesn't support such feature. Thus we need this tool to achieve this goal.

#### Implementation of LogPatternTool
This tool supports generating log patterns on the input dsl and index. It's implemented by
several steps:
1. Retrival [[${DOC_SIZE_FIELD}]] logs from index
2. Extract patterns for each retrieved log
 2.1 Find Pattern Field: If users provide parameter [[${PATTERN_FIELD}]], use it as the pattern
     field; Otherwise, find the string field with the longest length on the first log.
 2.2 Extract Pattern: If users provide parameter [[${PATTERN}]], compile it as a pattern;
     Otherwise, use [[${DEFAULT_IGNORED_CHARS}]]. It will remove all chars matching the pattern.
3. Group logs by their extracted patterns.
4. Find top N patterns with the largest sample log size.
5. For each found top N patterns, return [[${SAMPLE_LOG_SIZE}]] sample logs.

#### Example
```
POST _plugins/_ml/agents/V4pYKJIBf_90oTl_gmnp/_execute
{
    "parameters": {
       "selected_tools": ["LogPatternTool"],
       "top_n_pattern": "3",
       "sample_log_size": "20",
       "doc_size": 100,
        "input":  """
{"query":{"bool":{"filter":[{"range":{"timestamp":{"from":"1727172421713||-1h","to":"1727172421713","include_lower":true,"include_upper":true,"boost":1}}},{"term":{"response":{"value":"200","boost":1}}}],"adjust_pure_negative":true,"boost":1}}}
        """
    }
}
```
image

### Check List

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -68,7 +68,7 @@ public abstract class AbstractRetrieverTool implements Tool {
 
     protected abstract String getQueryBody(String queryText);
 
-    protected static Map<String, Object> processResponse(SearchHit hit) {
+    private static Map<String, Object> processResponse(SearchHit hit) {
         Map<String, Object> docContent = new HashMap<>();
         docContent.put("_index", hit.getIndex());
         docContent.put("_id", hit.getId());
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -89,8 +89,7 @@ public abstract class AbstractRetrieverTool implements Tool {
         searchSourceBuilder.parseXContent(queryParser);
         searchSourceBuilder.fetchSource(sourceFields, null);
         searchSourceBuilder.size(docSize);
-        SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(index);
-        return searchRequest;
+        return new SearchRequest().source(searchSourceBuilder).indices(parameters.getOrDefault(INDEX_FIELD, index));
     }
 
     @Override
--- a/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
+++ b/src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
@@ -68,7 +68,7 @@ public abstract class AbstractRetrieverTool implements Tool {
 
     protected abstract String getQueryBody(String queryText);
 
-    private static Map<String, Object> processResponse(SearchHit hit) {
+    protected static Map<String, Object> processResponse(SearchHit hit) {
         Map<String, Object> docContent = new HashMap<>();
         docContent.put("_index", hit.getIndex());
         docContent.put("_id", hit.getId());
@@ -77,7 +77,7 @@ public abstract class AbstractRetrieverTool implements Tool {
         return docContent;
     }
 
-    private <T> SearchRequest buildSearchRequest(Map<String, String> parameters) throws IOException {
+    protected <T> SearchRequest buildSearchRequest(Map<String, String> parameters) throws IOException {
         String question = parameters.get(INPUT_FIELD);
         if (StringUtils.isBlank(question)) {
             throw new IllegalArgumentException("[" + INPUT_FIELD + "] is null or empty, can not process it.");
--- a/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
+++ b/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
@@ -1,0 +1,255 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.agent.tools;
+
+import static org.opensearch.ml.common.utils.StringUtils.gson;
+
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.client.Client;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.ml.common.spi.tools.ToolAnnotation;
+import org.opensearch.search.SearchHit;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableSet;
+
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+
+/**
+ * This tool supports generating log patterns on the input dsl and index. It's implemented by
+ * several steps:
+ * 1. Retrival [[${DOC_SIZE_FIELD}]] logs from index
+ * 2. Extract patterns for each retrieved log
+ *  2.1 Find Pattern Field: If users provide parameter [[${PATTERN_FIELD}]], use it as the pattern
+ *      field; Otherwise, find the string field with the longest length on the first log.
+ *  2.2 Extract Pattern: If users provide parameter [[${PATTERN}]], compile it as a pattern;
+ *      Otherwise, use [[${DEFAULT_IGNORED_CHARS}]]. It will remove all chars matching the pattern.
+ * 3. Group logs by their extracted patterns.
+ * 4. Find top N patterns with the largest sample log size.
+ * 5. For each found top N patterns, return [[${SAMPLE_LOG_SIZE}]] sample logs.
+ */
+@Log4j2
+@Getter
+@Setter
+@ToolAnnotation(LogPatternTool.TYPE)
+public class LogPatternTool extends AbstractRetrieverTool {
+    public static final String TYPE = "LogPatternTool";
+
+    public static String DEFAULT_DESCRIPTION = "Log Pattern Tool";
+    public static final String TOP_N_PATTERN = "top_n_pattern";
+    public static final String SAMPLE_LOG_SIZE = "sample_log_size";
+    public static final String DSL_FIELD = "dsl";
+    public static final String PATTERN_FIELD = "pattern_field";
+    public static final String PATTERN = "pattern";
+    public static final int LOG_PATTERN_DEFAULT_DOC_SIZE = 1000;
+    public static final int DEFAULT_TOP_N_PATTERN = 3;
+    public static final int DEFAULT_SAMPLE_LOG_SIZE = 20;
+    private static final ImmutableSet<Character> DEFAULT_IGNORED_CHARS = ImmutableSet
+        .copyOf("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789".chars().mapToObj(c -> (char) c).toArray(Character[]::new));
+
+    private String name = TYPE;
+    private int topNPattern;
+    private int sampleLogSize;
+    @EqualsAndHashCode.Exclude
+    private Pattern pattern;
+
+    @Builder
+    public LogPatternTool(
+        Client client,
+        NamedXContentRegistry xContentRegistry,
+        int docSize,
+        int topNPattern,
+        int sampleLogSize,
+        String patternStr
+    ) {
+        super(client, xContentRegistry, null, null, docSize);
+        this.topNPattern = topNPattern;
+        this.sampleLogSize = sampleLogSize;
+        if (pattern != null)
+            this.pattern = Pattern.compile(patternStr);
+    }
+
+    @Override
+    protected String getQueryBody(String queryText) {
+        return queryText;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        super.index = parameters.get(INDEX_FIELD);
+        if (parameters.containsKey(DOC_SIZE_FIELD))
+            super.docSize = Integer.parseInt(parameters.get(DOC_SIZE_FIELD));
+        if (parameters.containsKey(TOP_N_PATTERN))
+            topNPattern = Integer.parseInt(parameters.get(TOP_N_PATTERN));
+        if (parameters.containsKey(SAMPLE_LOG_SIZE))
+            sampleLogSize = Integer.parseInt(parameters.get(SAMPLE_LOG_SIZE));
+        if (parameters.containsKey(PATTERN))
+            this.pattern = Pattern.compile((parameters.get(PATTERN)));
+
+        SearchRequest searchRequest;
+        try {
+            searchRequest = buildSearchRequest(parameters);
+        } catch (Exception e) {
+            log.error("Failed to build search request.", e);
+            listener.onFailure(e);
+            return;
+        }
+
+        ActionListener actionListener = ActionListener.<SearchResponse>wrap(r -> {
+            SearchHit[] hits = r.getHits().getHits();
+
+            if (hits != null && hits.length > 0) {
+                String patternField = parameters.containsKey(PATTERN_FIELD)
+                    ? parameters.get(PATTERN_FIELD)
+                    : findLongestField(hits[0].getSourceAsMap());
+                if (patternField == null) {
+                    listener.onResponse((T) "Pattern field is not set and this index doesn't contain any string field");
+                }
+                Map<String, List<Map<String, Object>>> patternGroups = new HashMap<>();
+                for (SearchHit hit : hits) {
+                    Map<String, Object> source = hit.getSourceAsMap();
+                    String pattern = extractPattern((String) source.getOrDefault(patternField, ""), this.pattern);
+                    List<Map<String, Object>> group = patternGroups.computeIfAbsent(pattern, k -> new ArrayList<>());
+                    group.add(source);
+                }
+                List<Map<String, Object>> sortedEntries = patternGroups
+                    .entrySet()
+                    .stream()
+                    .sorted(Comparator.comparingInt(entry -> -entry.getValue().size()))
+                    .limit(topNPattern)
+                    .map(
+                        entry -> Map
+                            .of(
+                                "total count",
+                                entry.getValue().size(),
+                                "pattern",
+                                entry.getKey(),
+                                "sample logs",
+                                entry.getValue().subList(0, Math.min(entry.getValue().size(), sampleLogSize))
+                            )
+                    )
+                    .toList();
+
+                listener
+                    .onResponse((T) AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> gson.toJson(sortedEntries)));
+            } else {
+                listener.onResponse((T) "Can not get any match from search result.");
+            }
+        }, e -> {
+            log.error("Failed to search index.", e);
+            listener.onFailure(e);
+        });
+        client.search(searchRequest, actionListener);
+    }
+
+    @VisibleForTesting
+    public static String extractPattern(String rawString, Pattern pattern) {
+        if (pattern != null)
+            return pattern.matcher(rawString).replaceAll("");
+        char[] chars = rawString.toCharArray();
+        int pos = 0;
+        for (int i = 0; i < chars.length; i++) {
+            if (!DEFAULT_IGNORED_CHARS.contains(chars[i])) {
+                chars[pos++] = chars[i];
+            }
+        }
+        return new String(chars, 0, pos);
+    }
+
+    @VisibleForTesting
+    public static String findLongestField(Map<String, Object> sampleLogSource) {
+        String longestField = null;
+        int maxLength = 0;
+
+        for (Map.Entry<String, Object> entry : sampleLogSource.entrySet()) {
+            Object value = entry.getValue();
+            if (value instanceof String) { // 确保值是字符串类型
+                String stringValue = (String) value;
+                int length = stringValue.length();
+                if (length > maxLength) {
+                    maxLength = length;
+                    longestField = entry.getKey();
+                }
+            }
+        }
+        return longestField;
+    }
+
+    @Override
+    public String getType() {
+        return TYPE;
+    }
+
+    public static class Factory extends AbstractRetrieverTool.Factory<LogPatternTool> {
+        private static LogPatternTool.Factory INSTANCE;
+
+        public static LogPatternTool.Factory getInstance() {
+            if (INSTANCE != null) {
+                return INSTANCE;
+            }
+            synchronized (LogPatternTool.class) {
+                if (INSTANCE != null) {
+                    return INSTANCE;
+                }
+                INSTANCE = new LogPatternTool.Factory();
+                return INSTANCE;
+            }
+        }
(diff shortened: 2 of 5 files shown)


src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    @VisibleForTesting
+    public static String findLongestField(Map<String, Object> sampleLogSource) {
+        String longestField = null;
+        int maxLength = 0;
+
+        for (Map.Entry<String, Object> entry : sampleLogSource.entrySet()) {
+            Object value = entry.getValue();
+            if (value instanceof String) { // 确保值是字符串类型
```
xluo-aws, line 184: Please translate the comment to English

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+            return;
+        }
+
+        ActionListener actionListener = ActionListener.<SearchResponse>wrap(r -> {
+            SearchHit[] hits = r.getHits().getHits();
+
+            if (hits != null && hits.length > 0) {
+                String patternField = parameters.containsKey(PATTERN_FIELD)
+                    ? parameters.get(PATTERN_FIELD)
+                    : findLongestField(hits[0].getSourceAsMap());
```
xluo-aws, line 122: Should we stop if longestField is fewer than maybe 30 chars or continue to search for 10 more records in case the first one has outlier data?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+        NamedXContentRegistry xContentRegistry,
+        int docSize,
+        int topNPattern,
+        int sampleLogSize,
+        String patternStr
+    ) {
+        super(client, xContentRegistry, null, null, docSize);
+        this.topNPattern = topNPattern;
+        this.sampleLogSize = sampleLogSize;
+        if (pattern != null)
```
xluo-aws, line 86: Should check patternStr instead of pattern

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+
+    @Override
+    protected String getQueryBody(String queryText) {
+        return queryText;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        super.index = parameters.get(INDEX_FIELD);
+        if (parameters.containsKey(DOC_SIZE_FIELD))
```
xluo-aws, line 98: Use parameters.getOrDefault?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+@Getter
+@Setter
+@ToolAnnotation(LogPatternTool.TYPE)
+public class LogPatternTool extends AbstractRetrieverTool {
+    public static final String TYPE = "LogPatternTool";
+
+    public static String DEFAULT_DESCRIPTION = "Log Pattern Tool";
+    public static final String TOP_N_PATTERN = "top_n_pattern";
+    public static final String SAMPLE_LOG_SIZE = "sample_log_size";
+    public static final String DSL_FIELD = "dsl";
```
yuye-aws, line 59: Why this field is not used?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+    @Override
+    protected String getQueryBody(String queryText) {
+        return queryText;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        super.index = parameters.get(INDEX_FIELD);
+        if (parameters.containsKey(DOC_SIZE_FIELD))
+            super.docSize = Integer.parseInt(parameters.get(DOC_SIZE_FIELD));
```
yuye-aws, line 99: `Integer.parseInt` could throw `NumberFormatException`. Please add try catch for error handling and better exception message.

Also, please add some validation that these size are positive integer.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+        if (parameters.containsKey(TOP_N_PATTERN))
+            topNPattern = Integer.parseInt(parameters.get(TOP_N_PATTERN));
+        if (parameters.containsKey(SAMPLE_LOG_SIZE))
+            sampleLogSize = Integer.parseInt(parameters.get(SAMPLE_LOG_SIZE));
+        if (parameters.containsKey(PATTERN))
+            this.pattern = Pattern.compile((parameters.get(PATTERN)));
```
yuye-aws, line 105: Try-catch

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+        for (int i = 0; i < chars.length; i++) {
+            if (!DEFAULT_IGNORED_CHARS.contains(chars[i])) {
+                chars[pos++] = chars[i];
+            }
+        }
+        return new String(chars, 0, pos);
+    }
+
+    @VisibleForTesting
+    public static String findLongestField(Map<String, Object> sampleLogSource) {
```
yuye-aws, line 178: This method is already visible for testing. Should it be private?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+        }, e -> {
+            log.error("Failed to search index.", e);
+            listener.onFailure(e);
+        });
+        client.search(searchRequest, actionListener);
+    }
+
+    @VisibleForTesting
+    public static String extractPattern(String rawString, Pattern pattern) {
```
yuye-aws, line 164: plz change to private method

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+            int docSize = params.containsKey(DOC_SIZE_FIELD)
+                ? Integer.parseInt((String) params.get(DOC_SIZE_FIELD))
+                : LOG_PATTERN_DEFAULT_DOC_SIZE;
+            int topNPattern = params.containsKey(TOP_N_PATTERN)
+                ? Integer.parseInt((String) params.get(TOP_N_PATTERN))
+                : DEFAULT_TOP_N_PATTERN;
+            int sampleLogSize = params.containsKey(SAMPLE_LOG_SIZE)
+                ? Integer.parseInt((String) params.get(SAMPLE_LOG_SIZE))
+                : DEFAULT_SAMPLE_LOG_SIZE;
+            String patternStr = params.containsKey(PATTERN) ? (String) params.get(PATTERN) : null;
```
yuye-aws, line 228: You can consider factory as a wrapper to create tools. I would prefer add the parameter validation logic to during the construction of the `LogPatternTool`

yuye-aws (commented):
You can implement the ut and it after resolving my comments

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: It just imitates the logic of Discover log pattern. 

It sounds good to avoid outliner, but it could also make the logic more complex and inconsistent with Discover.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: Will delete it. 

At the beginning, I suppose to use this field to pass in the `DSL` parameter and build the request myself. But I find I can reuse the fucntion  `buildSearchRequest` in the base class `AbstractRetrieverTool`, and it only accepts field `input`.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: You're right. I spot that problem in `VectorDBTool` and `NeuralSparseSearchTool` as well, so I keep consistent with them here. But I think it's great to raise the bar!

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: The idea why I didn't using getOrDefault here is that we always have `docSize` set when executing. Nor by the parameters we pass when creating this tool, or the default value.

So if users pass parameters when executing, it's actually always an operation of overriding parameters. Thus, in my opinion, I think checking key instead of using `getOrDefault` will make it more readable by  indicating that meaning, and also avoid second evaluation if user doesn't pass parameter.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: Actually the annotation @VisibleForTesting doesn't take effect here, that's why I change it from private to public here as I haven't got time to investigate why.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: Just keep the same style with `VectorDBTool` and `NeuralSparseSearchTool` which all extend the same parent class `AbstractRetrieverTool`.  
 
Will do that change since it makes more sense.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
xluo-aws: The above suggestion is not ideal, just want to trigger some discussion. It's ok to leave it as it is if we don't have better idea.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
yuye-aws: You can simply declare like `static String .....`

yuye-aws (commented):
Did not see the PPL, but it makes sense to use DSL for search. Would you like to check whether the code can be reused with search index tool? In search index tool, it has specialized operations on connector, model and model_group index.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+    public static String extractPattern(String rawString, Pattern pattern) {
+        if (pattern != null)
+            return pattern.matcher(rawString).replaceAll("");
+        char[] chars = rawString.toCharArray();
+        int pos = 0;
+        for (int i = 0; i < chars.length; i++) {
+            if (!DEFAULT_IGNORED_CHARS.contains(chars[i])) {
+                chars[pos++] = chars[i];
+            }
+        }
```
yuye-aws, line 173: For better readability, please try regex: `[a-zA-Z0-9]`

src/main/java/org/opensearch/agent/tools/AbstractRetrieverTool.java
```diff
@@ -68,7 +68,7 @@ protected AbstractRetrieverTool(
 
     protected abstract String getQueryBody(String queryText);
 
-    private static Map<String, Object> processResponse(SearchHit hit) {
+    protected static Map<String, Object> processResponse(SearchHit hit) {
```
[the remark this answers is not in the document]
qianheng-aws: Addressed. I suppose to override it at the beginning but found  I need to override the whole run function.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
yuye-aws, line 122: What if the longest field is missing in `hits[0]` ? Shall we iterate though all the hits?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+        Client client,
+        NamedXContentRegistry xContentRegistry,
+        int docSize,
+        int topNPattern,
+        int sampleLogSize,
+        String patternStr
+    ) {
+        super(client, xContentRegistry, null, null, docSize);
+        this.topNPattern = topNPattern;
+        this.sampleLogSize = sampleLogSize;
```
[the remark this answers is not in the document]
qianheng-aws: run time parameters I think.  But I think users can set them as well when creating this tool, and it should support override them in runtime.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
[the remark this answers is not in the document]
qianheng-aws: Do you mean it already has the logic to do update itself?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,255 @@
...
+        SearchRequest searchRequest;
+        try {
+            searchRequest = buildSearchRequest(parameters);
+        } catch (Exception e) {
+            log.error("Failed to build search request.", e);
+            listener.onFailure(e);
+            return;
+        }
+
+        ActionListener actionListener = ActionListener.<SearchResponse>wrap(r -> {
```
yuye-aws, line 116: Prefer `ActionListener<SearchResponse> actionListener = ActionListener.wrap`

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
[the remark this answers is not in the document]
yuye-aws: I mean `parameters` is about runtime parameters. It only affects within the run method.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: Yeah, that's possible. And also, in some cases, the longest field in the first log maybe not the most longest overall. 

But I think firstly we just imitate what current log pattern feature do, and do enhancement later. Better to discuss with Observability team as well. You can check the current implementation here [Code Ref](https://github.com/opensearch-project/dashboards-observability/blob/279463ad0c8e00780bf57f85e6d2bf30e8c7c93d/public/components/event_analytics/hooks/use_fetch_patterns.ts#L137C2-L164C6)

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: For this version, I think we just need to copy the code from sql plugin without doing too much change. So developer can better track the source and diff.

What you suggested could be an enhancement in a separate PR.

qianheng-aws:
> Did not see the PPL, but it makes sense to use DSL for search. Would you like to check whether the code can be reused with search index tool? In search index tool, it has specialized operations on connector, model and model_group index.

LogPatternTool aims to generate log pattern for DSL. PPL already supports `pattern` syntax itself.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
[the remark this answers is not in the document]
qianheng-aws: Got it. Because AbstractRetrieverTool has stored tool parameters in its fields,  maybe we shouldn't update the tool  parameters since it's singleton instance.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
yuye-aws: Sure

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @Override
+    protected String getQueryBody(String queryText) {
+        return queryText;
+    }
+
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        // LogPatternTool needs to pass index and input as parameter in runtime. See [[${validate}]]
+        super.index = parameters.get(INDEX_FIELD);
```
yuye-aws, line 101: should not update this.index with run time parameters

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,294 @@
...
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException(
+                LoggerMessageFormat.format("Invalid value {} for parameter {}, it should be a number", params.get(paramName), paramName)
+            );
+        }
+        return value;
+    }
+
+    private static void checkPositive(int value, String paramName) {
+        if (value < 0) {
```
yuye-aws, line 239: value <= 0

yuye-aws (commented):
Good job @qianheng-aws ! You can go on with ut and it after resolving these two mistakes.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: It make < 0 on purpose as I found many places in OpenSearch using "<0" for checking positive. See some examples below:

1. https://github.com/opensearch-project/OpenSearch/blob/e8b02c9545d7908ae1a3ea2d109c06ff3e0c07a5/server/src/main/java/org/opensearch/index/reindex/BulkByScrollTask.java#L917

2. org.opensearch.jobscheduler.repackage.com.cronutils.utils.Preconditions::checkPositive
```
  public static void checkPositive(int value, Object errorMessage) {
    if (value < 0) {
      throw new IllegalArgumentException(String.valueOf(errorMessage));
    }
  }
```

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: It seems tricky but cannot find better solutions since `AbstractRetrieverTool` take `index` as tool parameter. But I think it should be runtime parameter

Do you have any suggestions?

qianheng-aws: Does it make sense to make minor change in `AbstractRetrieverTool#buildSearchRequest`:

```
// Before
        SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(index);

// After
        SearchRequest searchRequest = new SearchRequest().source(searchSourceBuilder).indices(parameters.getOrDefault(INDEX_FIELD, index));
```

Then I don't need to change the field `super.index`.

yuye-aws: Yeah, it totally makes sense.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
yuye-aws: Then rename it to `checkNonNegative`

qianheng-aws: I wonder if it's a habit to check '< 0' with function name `checkPostive`.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,293 @@
...
+ * 5. For each found top N patterns, return [[${SAMPLE_LOG_SIZE}]] sample logs.
+ */
+@Log4j2
+@Getter
+@Setter
+@ToolAnnotation(LogPatternTool.TYPE)
+public class LogPatternTool extends AbstractRetrieverTool {
+    public static final String TYPE = "LogPatternTool";
+
+    public static String DEFAULT_DESCRIPTION = "Log Pattern Tool";
```
zane-neo, line 58: This can be final

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -0,0 +1,293 @@
...
+
+        ActionListener<SearchResponse> actionListener = ActionListener.wrap(r -> {
+            SearchHit[] hits = r.getHits().getHits();
+
+            if (hits != null && hits.length > 0) {
+                String patternField = parameters.containsKey(PATTERN_FIELD)
+                    ? parameters.get(PATTERN_FIELD)
+                    : findLongestField(hits[0].getSourceAsMap());
+                if (patternField == null) {
+                    listener.onResponse((T) "Pattern field is not set and this index doesn't contain any string field");
```
zane-neo, line 121: Do we need a return here? Or it's as expected to continue even patternField is null, if yes, there's several places needs to handle NPE in following code.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
zane-neo: It's consistent to check `<0` in method `checkPositive`, but in our case, 0 doesn't make sense either for top_n or sample_count, right? And in this case, we should let user know their configuration issue.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
qianheng-aws: Yes, miss return.

zane-neo:
It's better to add ITs to cover the code, is there any concern on adding ITs?

qianheng-aws:
> It's better to add ITs to cover the code, is there any concern on adding ITs?

I'm going to add one for this PR.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
xinyual: Can we add one more check to see whether parseField is one of the columns inside the index? Sometimes customer could type wrong parse field name, in the PPL, it will raise an error to show the parse field doesn't exist. We could do the same logic.

zane-neo (approved):

yuye-aws:
@qianheng-aws Please raise another PR with ut and it.

PR #452, opened 2024-10-23
Support pass prompt to CreateAlertTool

qianheng-aws:
### Description
This PR enhance the CreateAlertTool with 2 features:
- support pass prompt as parameter to override the default prompt.
- use CLAUDE  as its default model type if model_type in parameters is empty or unsupported.

### Related Issues
Resolves https://github.com/opensearch-project/skills/issues/447

### Check List
- [x] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/skills/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #457, opened 2024-11-01
CreateAnomalyDetectorTool supports empty model_type

gaobinlong:
### Description
Same to the PPLTool, CreateAnomalyDetectorTool should also support empty model_type parameter, rather than throwing an `IllegalArgumentException`, we need to make the behavior consistent between the tools.

### Related Issues
No issue.

### Check List

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
+++ b/src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
@@ -431,8 +431,8 @@ public class CreateAnomalyDetectorTool implements Tool {
             String modelType = (String) map.getOrDefault("model_type", ModelType.CLAUDE.toString());
             if (modelType.isEmpty()) {
                 modelType = ModelType.CLAUDE.toString();
-            }
-            if (!ModelType.OPENAI.toString().equalsIgnoreCase(modelType) && !ModelType.CLAUDE.toString().equalsIgnoreCase(modelType)) {
+            } else if (!ModelType.OPENAI.toString().equalsIgnoreCase(modelType)
+                && !ModelType.CLAUDE.toString().equalsIgnoreCase(modelType)) {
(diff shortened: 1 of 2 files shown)


yuye-aws (approved):
Looks good to me.

yuye-aws:
[Flakey tests](https://github.com/opensearch-project/skills/actions/runs/11626676324/job/32378711529?pr=457) coming from RAG Tool IT. Can @mingshl take a look?

yuye-aws:
@gaobinlong Can you check the comment here https://github.com/opensearch-project/skills/pull/452#discussion_r1828681195? You can update the PR after Heng's PR gets merged.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -429,7 +429,11 @@ public CreateAnomalyDetectorTool create(Map<String, Object> map) {
                 throw new IllegalArgumentException("model_id cannot be empty.");
             }
             String modelType = (String) map.getOrDefault("model_type", ModelType.CLAUDE.toString());
-            if (!ModelType.OPENAI.toString().equalsIgnoreCase(modelType) && !ModelType.CLAUDE.toString().equalsIgnoreCase(modelType)) {
+            // if model type is empty, use the default value
+            if (modelType.isEmpty()) {
```
zane-neo, line 433: Not understand why this change as in line 431 there's already a default value which should suggest `modelType` is not empty, isn't it?

gaobinlong: No, the default value takes effect only if the key `model_type` doesn't exist in the map or its value is null.

zane-neo: Got it, this is to handle empty string case instead of null value.

zane-neo (approved):

PR #474, opened 2025-01-02
Log pattern tool improvement

songkant-aws:
### Description
Improve log pattern tool with cutting edge log pattern parser by introducing Brain algorithm mentioned in this RFC: https://github.com/opensearch-project/OpenSearch/issues/16627

### Check List

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
+++ b/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
@@ -229,7 +229,7 @@ public class LogPatternTool extends AbstractRetrieverTool {
         int topNPattern = parameters.containsKey(TOP_N_PATTERN) ? getPositiveInteger(parameters, TOP_N_PATTERN) : this.topNPattern;
         int sampleLogSize = parameters.containsKey(SAMPLE_LOG_SIZE) ? getPositiveInteger(parameters, SAMPLE_LOG_SIZE) : this.sampleLogSize;
 
-        Map<String, List<String>> logPatternMap = logParser.parseAllLogPatterns(logMessages, List.of());
+        Map<String, List<String>> logPatternMap = logParser.parseAllLogPatterns(logMessages);
 
         return logPatternMap
             .entrySet()
--- a/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
+++ b/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
@@ -15,6 +15,7 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.IntStream;
@@ -302,7 +303,7 @@ public class LogPatternTool extends AbstractRetrieverTool {
          * i.e. TIMESTAMPADD(DAY, -1, '2025-01-01 00:00:00') is different from TIMESTAMPADD(day, -1, '2025-01-01 00:00:00')
          * The latter one is not parsed well by PPLService.
          */
-        int idx = normPPL.toUpperCase().indexOf("| STATS");
+        int idx = normPPL.toUpperCase(Locale.ROOT).indexOf("| STATS");
         return idx != -1 ? normPPL.substring(0, idx).trim() : ppl;
     }
 
--- a/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
+++ b/src/main/java/org/opensearch/agent/tools/LogPatternTool.java
@@ -8,6 +8,7 @@ package org.opensearch.agent.tools;
 import static org.opensearch.ml.common.utils.StringUtils.gson;
 
 import java.security.AccessController;
+import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -15,7 +16,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
+import java.util.function.Function;
 import java.util.stream.IntStream;
 
 import org.apache.commons.lang3.StringUtils;
@@ -27,6 +28,7 @@ import org.opensearch.agent.tools.utils.ToolHelper;
 import org.opensearch.client.Client;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.common.logging.LoggerMessageFormat;
+import org.opensearch.core.common.util.CollectionUtils;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.ml.common.spi.tools.ToolAnnotation;
 import org.opensearch.search.SearchHit;
@@ -36,8 +38,6 @@ import org.opensearch.sql.plugin.transport.TransportPPLQueryResponse;
 import org.opensearch.sql.ppl.domain.PPLQueryRequest;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import com.google.gson.reflect.TypeToken;
 
 import lombok.Builder;
@@ -49,7 +49,7 @@ import lombok.extern.log4j.Log4j2;
  * This tool supports generating log patterns on the input dsl and index. It's implemented by
  * several steps:
  * 1. Retrival [[${DOC_SIZE_FIELD}]] logs from index by either dsl or ppl query
- * 2. Extract patterns for each retrieved log: If users provide parameter [[${PATTERN_FIELD}]], use it as the pattern
+ * 2. Extract patterns for input logs: If users provide parameter [[${PATTERN_FIELD}]], use it as the pattern
  *      field; Otherwise, find the string field with the longest length on the first log.
  * 3. Group logs by their extracted patterns.
  * 4. Find top N patterns with the largest sample log size.
@@ -95,9 +95,9 @@ public class LogPatternTool extends AbstractRetrieverTool {
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        String question = parameters.get(INPUT_FIELD);
+        String dsl = parameters.get(INPUT_FIELD);
         String ppl = parameters.get(PPL_FIELD);
-        if (!StringUtils.isBlank(question)) {
+        if (!StringUtils.isBlank(dsl)) {
             SearchRequest searchRequest;
             try {
                 searchRequest = buildSearchRequest(parameters);
@@ -110,24 +110,15 @@ public class LogPatternTool extends AbstractRetrieverTool {
             ActionListener<SearchResponse> actionListener = ActionListener.wrap(r -> {
                 SearchHit[] hits = r.getHits().getHits();
 
-                if (hits != null && hits.length > 0) {
+                if (!CollectionUtils.isEmpty(hits)) {
                     Map<String, Object> firstLogSource = hits[0].getSourceAsMap();
-                    String patternField = parameters.containsKey(PATTERN_FIELD)
-                        ? parameters.get(PATTERN_FIELD)
-                        : findLongestField(firstLogSource);
-                    validatePatternFieldAndFirstLogSource(parameters, patternField, firstLogSource);
 
-                    List<String> logMessages = Arrays.stream(hits).map(hit -> {
+                    Function<String, List<String>> logMessagesProvider = (String patternField) -> Arrays.stream(hits).map(hit -> {
                         Map<String, Object> source = hit.getSourceAsMap();
                         return (String) source.getOrDefault(patternField, "");
                     }).toList();
 
-                    List<Map<String, Object>> sortedEntries = getTopNLogPatterns(parameters, logMessages);
-
-                    listener
-                        .onResponse(
-                            (T) AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> gson.toJson(sortedEntries))
-                        );
+                    onResponseSortedLogPatterns(parameters, listener, firstLogSource, logMessagesProvider);
                 } else {
                     listener.onResponse((T) "Can not get any match from search result.");
                 }
@@ -138,8 +129,7 @@ public class LogPatternTool extends AbstractRetrieverTool {
             client.search(searchRequest, actionListener);
         } else if (!StringUtils.isBlank(ppl)) {
             String prunedPPL = removePPLAggregations(ppl);
-            JSONObject jsonContent = new JSONObject(ImmutableMap.of("query", prunedPPL));
-            PPLQueryRequest pplQueryRequest = new PPLQueryRequest(prunedPPL, jsonContent, null, "jdbc");
+            PPLQueryRequest pplQueryRequest = new PPLQueryRequest(prunedPPL, null, null, "jdbc");
             TransportPPLQueryRequest transportPPLQueryRequest = new TransportPPLQueryRequest(pplQueryRequest);
 
             ActionListener<TransportPPLQueryResponse> actionListener = ActionListener.wrap(r -> {
@@ -157,27 +147,15 @@ public class LogPatternTool extends AbstractRetrieverTool {
                         .filter(i -> schema.get(i) != null && !StringUtils.isBlank(schema.get(i).get(PPL_SCHEMA_NAME)))
                         .forEach(i -> firstLogSource.put(schema.get(i).get(PPL_SCHEMA_NAME), firstDataRow.get(i)));
 
-                    String patternField = parameters.containsKey(PATTERN_FIELD)
-                        ? parameters.get(PATTERN_FIELD)
-                        : findLongestField(firstLogSource);
-                    validatePatternFieldAndFirstLogSource(parameters, patternField, firstLogSource);
-
-                    List<String> logMessages = new ArrayList<>();
-                    Optional<Integer> fieldIndex = IntStream
+                    Function<String, List<String>> logMessagesProvider = (String patternField) -> IntStream
                         .range(0, schema.size())
                         .boxed()
                         .filter(i -> schema.get(i) != null && patternField.equals(schema.get(i).get(PPL_SCHEMA_NAME)))
-                        .findFirst();
-                    if (fieldIndex.isPresent()) {
-                        logMessages = dataRows.stream().map(dataRow -> (String) dataRow.get(fieldIndex.get())).toList();
-                    }
+                        .findFirst()
+                        .map(fieldIndex -> dataRows.stream().map(dataRow -> (String) dataRow.get(fieldIndex)).toList())
+                        .orElseGet(ArrayList::new);
 
-                    List<Map<String, Object>> sortedEntries = getTopNLogPatterns(parameters, logMessages);
-
-                    listener
-                        .onResponse(
-                            (T) AccessController.doPrivileged((PrivilegedExceptionAction<String>) () -> gson.toJson(sortedEntries))
-                        );
+                    onResponseSortedLogPatterns(parameters, listener, firstLogSource, logMessagesProvider);
                 } else {
                     listener.onResponse((T) "Can not get any data row from ppl response.");
                 }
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
     }
 
     @Override
     public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
-        int topNPattern = parameters.containsKey(TOP_N_PATTERN) ? getPositiveInteger(parameters, TOP_N_PATTERN) : this.topNPattern;
-        int sampleLogSize = parameters.containsKey(SAMPLE_LOG_SIZE) ? getPositiveInteger(parameters, SAMPLE_LOG_SIZE) : this.sampleLogSize;
-        Pattern pattern = parameters.containsKey(PATTERN) ? Pattern.compile(parameters.get(PATTERN)) : this.pattern;
+        String question = parameters.get(INPUT_FIELD);
+        String ppl = parameters.get(PPL_FIELD);
+        if (!StringUtils.isBlank(question)) {
```
gaobinlong, line 100: Maybe it's better to use the common method `Strings.isNullOrEmpty()` in OS-core instead?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
-                            )
-                    );
-                }
-                Map<String, List<Map<String, Object>>> patternGroups = new HashMap<>();
-                for (SearchHit hit : hits) {
-                    Map<String, Object> source = hit.getSourceAsMap();
-                    String patternValue = extractPattern((String) source.getOrDefault(patternField, ""), pattern);
-                    List<Map<String, Object>> group = patternGroups.computeIfAbsent(patternValue, k -> new ArrayList<>());
-                    group.add(source);
+                if (hits != null && hits.length > 0) {
```
gaobinlong, line 113: We have `CollectionUtils.isEmpty()` method in OS-core can be used here.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
-                    )
-                    .toList();
+            }, e -> {
+                log.error("Failed to search index.", e);
+                listener.onFailure(e);
+            });
+            client.search(searchRequest, actionListener);
+        } else if (!StringUtils.isBlank(ppl)) {
+            String prunedPPL = removePPLAggregations(ppl);
+            JSONObject jsonContent = new JSONObject(ImmutableMap.of("query", prunedPPL));
```
gaobinlong, line 141: Why not using Map.of() in JDK?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
-        int pos = 0;
-        for (int i = 0; i < chars.length; i++) {
-            if (!DEFAULT_IGNORED_CHARS.contains(chars[i])) {
-                chars[pos++] = chars[i];
-            }
+    List<Map<String, Object>> getTopNLogPatterns(Map<String, String> parameters, List<String> logMessages) {
+        int topNPattern = parameters.containsKey(TOP_N_PATTERN) ? getPositiveInteger(parameters, TOP_N_PATTERN) : this.topNPattern;
+        int sampleLogSize = parameters.containsKey(SAMPLE_LOG_SIZE) ? getPositiveInteger(parameters, SAMPLE_LOG_SIZE) : this.sampleLogSize;
+
+        Map<String, List<String>> logPatternMap = logParser.parseAllLogPatterns(logMessages, ImmutableList.of());
```
gaobinlong, line 200: Maybe List.of() is better.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -233,8 +282,35 @@ public String getType() {
...
+        /*
+         * Remove all following query starting with stats as they rely on aggregation results.
+         * We don't convert ppl string to upper or lower case because some enum parameters of functions are case-sensitive.
+         * i.e. TIMESTAMPADD(DAY, -1, '2025-01-01 00:00:00') is different from TIMESTAMPADD(day, -1, '2025-01-01 00:00:00')
+         * The latter one is not parsed well by PPLService.
+         */
+        int idx = Math
+            .min(
+                !normPPL.contains("| stats") ? Integer.MAX_VALUE : normPPL.indexOf("| stats"),
+                !normPPL.contains("| STATS") ? Integer.MAX_VALUE : normPPL.indexOf("| STATS")
```
xluo-aws, line 311: Can we use normPPL.toUpperCase.indexof(" | STATS")

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
xluo-aws, line 100: buildSearchRequest will do the check anyway, no need to add this logic

xluo-aws, line 100: Question is DSL?

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
-                    .toList();
+            }, e -> {
+                log.error("Failed to search index.", e);
+                listener.onFailure(e);
+            });
+            client.search(searchRequest, actionListener);
+        } else if (!StringUtils.isBlank(ppl)) {
+            String prunedPPL = removePPLAggregations(ppl);
+            JSONObject jsonContent = new JSONObject(ImmutableMap.of("query", prunedPPL));
+            PPLQueryRequest pplQueryRequest = new PPLQueryRequest(prunedPPL, jsonContent, null, "jdbc");
```
xluo-aws, line 142: jsonContext also has PPL info? Curious why we need it in both parameters.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: I kept the previous code style to use google guava collection. I'm OK to use JDK APIs.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: I kept the previous code style to use google guava collection. I'm OK to use JDK APIs.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: Changed to OS-core Strings util function

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: Changed to OS-core Strings util function

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: I'm not quite sure if there is other case of function or enums preferring lower case. That's why I made such code change. It's certain that '| stats' is equivalent to '| STATS'.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: Yes, question is dsl actually. Maybe it's a bad variable naming. I changed to 'dsl'

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
songkant-aws: Yeah, that's odd. I followed the code pattern of executing PPLRequest in PPLTool. 

But I just looked the code of TransportPPLQueryAction. It doesn't seem to use it in following PPLService of transforming to execution plan.

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+        Map<Long, Integer> occurrences = new HashMap<>();
+        for (int i = 0; i < tokens.size() - 1; i++) {
+            String tokenKey = String.format(Locale.ROOT, POSITIONED_TOKEN_KEY_FORMAT, i, tokens.get(i));
+            Long tokenFreq = tokenFreqMap.get(tokenKey);
+            occurrences.put(tokenFreq, occurrences.getOrDefault(tokenFreq, 0) + 1);
+        }
+        return occurrences;
+    }
+
+    private List<Map.Entry<Long, Integer>> getSortedWordCombinations(Map<Long, Integer> occurrences) {
```
lyx-aws, line 298: Maybe consider to wrap the <Long, Integer> and related operations into a class for better encapsulation? Then we can implement Comparable interface for the sort operation and use `Collections.sort()`

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -59,145 +66,187 @@ public class LogPatternTool extends AbstractRetrieverTool {
...
-                }
-                Map<String, List<Map<String, Object>>> patternGroups = new HashMap<>();
-                for (SearchHit hit : hits) {
-                    Map<String, Object> source = hit.getSourceAsMap();
-                    String patternValue = extractPattern((String) source.getOrDefault(patternField, ""), pattern);
-                    List<Map<String, Object>> group = patternGroups.computeIfAbsent(patternValue, k -> new ArrayList<>());
-                    group.add(source);
+                if (hits != null && hits.length > 0) {
+                    Map<String, Object> firstLogSource = hits[0].getSourceAsMap();
+                    String patternField = parameters.containsKey(PATTERN_FIELD)
```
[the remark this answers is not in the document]
songkant-aws: Cool, done.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
xluo-aws: All test cases for PPLRequest have jsonContent as null

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
xluo-aws: Don't need remove aggregation for DSL?

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+/**
+ * Log parser Brain algorithm implementation. See: https://ieeexplore.ieee.org/document/10109145
+ */
+public class BrainLogParser {
+
+    private static final String VARIABLE_DENOTER = "<*>";
+    private static final Map<Pattern, String> DEFAULT_FILTER_PATTERN_VARIABLE_MAP = new LinkedHashMap<>();
+    static {
+        // IP
+        DEFAULT_FILTER_PATTERN_VARIABLE_MAP.put(Pattern.compile("(/|)([0-9]+\\.){3}[0-9]+(:[0-9]+|)(:|)"), "<*IP*>");
```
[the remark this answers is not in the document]
xluo-aws: For IP4, ^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$ is a more strict per Q

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+        // Simple ISO date and time
+        DEFAULT_FILTER_PATTERN_VARIABLE_MAP
+            .put(Pattern.compile("(\\d{4}-\\d{2}-\\d{2})[T ]?(\\d{2}:\\d{2}:\\d{2})(\\.\\d{3})?(Z|([+-]\\d{2}:?\\d{2}))?"), "<*DATETIME*>");
```
xluo-aws, line 35: Here is a more general one 
(?:
    # Common date patterns
    (?:\d{4}[-/.]\d{2}[-/.]\d{2})|    # YYYY-MM-DD, YYYY/MM/DD
    (?:\d{2}[-/.]\d{2}[-/.]\d{4})|    # DD-MM-YYYY, MM/DD/YYYY
    (?:[A-Za-z]{3}\s+\d{1,2})|        # Month abbreviation + day (Jan 23)
    (?:\d{1,2}\s+[A-Za-z]{3})         # Day + month abbreviation (23 Jan)
)
(?:
    # Time patterns
    (?:[\sT]?\d{2}[:]\d{2}(?:[:]\d{2})?)?  # HH:MM:SS or HH:MM
    (?:\.\d+)?                               # milliseconds
    (?:Z|[+-]\d{2}:?\d{2})?                 # timezone
    (?:\s?(?:AM|PM|am|pm))?                 # AM/PM
)?

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+        // Hex Decimal, letters followed by digits, float numbers
+        DEFAULT_FILTER_PATTERN_VARIABLE_MAP
+            .put(Pattern.compile("((0x|0X)[0-9a-fA-F]+)|[a-zA-Z]+\\d+|([+-]?(\\d+(\\.\\d*)?|\\.\\d+))"), VARIABLE_DENOTER);
+        // generic number surrounded by non-alphanumeric
+        DEFAULT_FILTER_PATTERN_VARIABLE_MAP.put(Pattern.compile("(?<=[^A-Za-z0-9])(-?\\+?\\d+)(?=[^A-Za-z0-9])|[0-9]+$"), VARIABLE_DENOTER);
```
xluo-aws, line 40: is this a common pattern to detect variable?

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+        int variableCountThreshold,
+        float thresholdPercentage,
+        Map<Pattern, String> filterPatternVariableMap,
+        List<String> delimiters
+    ) {
+        this.tokenFreqMap = new HashMap<>();
+        this.groupTokenSetMap = new HashMap<>();
+        this.logIdGroupCandidateMap = new HashMap<>();
+        this.variableCountThreshold = variableCountThreshold;
+        if (thresholdPercentage < 0.0f || thresholdPercentage > 1.0f) {
```
xluo-aws, line 115: Move to 111 . Fail fast

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
songkant-aws: That's fair. I see this pair is used in other places. So make it a record class(a kind of new data class in newer JDK version?) and implements Comparable

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
songkant-aws: Done

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+    /**
+     * Preprocess all lines of log messages with logId list. Empty logId list is allowed as the index within
+     * the list will be logId by default
+     * @param logMessages list of log messages
+     * @param logIds list of logIds corresponded to log message
+     * @return list of token lists
+     */
+    public List<List<String>> preprocessAllLogs(List<String> logMessages, List<String> logIds) {
+        List<List<String>> preprocessedLogs = new ArrayList<>();
+        int size = logIds.isEmpty() ? logMessages.size() : Math.min(logMessages.size(), logIds.size());
```
xluo-aws, line 169: Do we have a real case that logIds will be passed along with logMessage? The concern is if they don't match each other, the parse result will not match as well. Either remove the logIds parameter if it's not required or throw except if it doesn't match logMessages.

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+    public Map<String, String> getLogIdGroupCandidateMap() {
+        return this.logIdGroupCandidateMap;
+    }
+
+    private Map<Long, Integer> getWordOccurrences(List<String> tokens) {
+        Map<Long, Integer> occurrences = new HashMap<>();
+        for (int i = 0; i < tokens.size() - 1; i++) {
+            String tokenKey = String.format(Locale.ROOT, POSITIONED_TOKEN_KEY_FORMAT, i, tokens.get(i));
+            Long tokenFreq = tokenFreqMap.get(tokenKey);
+            occurrences.put(tokenFreq, occurrences.getOrDefault(tokenFreq, 0) + 1);
```
xluo-aws, line 298: Still not quite understand this logic, if words at different position have the same frequency, does it mean they always appear in the same line of log? If we can't have this assumption, how can we consider them as the longest combination?

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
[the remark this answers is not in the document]
xluo-aws: This can be enhanced later.

xluo-aws (approved):

lyx-aws (approved):

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
songkant-aws: We don't have a real case yet. I'm OK with removing it in current skills repo. Since this algorithm does two passes to iterate original logs, it was originally considered to use logIds to identify in scope searched doc in second pass search if we want to apply distributed search. But that plan is not finalized yet, so I remove the logIds input for now.

src/main/java/org/opensearch/agent/tools/LogPatternTool.java
```diff
@@ -233,8 +204,106 @@ public String getType() {
...
+
+    private String removeDSLAggregations(String dsl) {
+        JSONObject dslObj = new JSONObject(dsl);
+        // DSL request is a json blob. Aggregations usually have keys 'aggs' or 'aggregations'
+        dslObj.remove("aggs");
+        dslObj.remove("aggregations");
+        return dslObj.toString();
+    }
+
+    private String removePPLAggregations(String ppl) {
```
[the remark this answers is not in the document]
songkant-aws: Yes, for simplicity, we don't expect such cases. If there is one in future, we could revisit this assumption. For now, we only want to query on original raw logs with filters applied on them, not aggregation results.

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
```
xluo-aws, line 294: token with same frequency may not always appear in the same line of log, right?

src/main/java/org/opensearch/agent/tools/utils/BrainLogParser.java
```diff
@@ -0,0 +1,336 @@
...
+        List<List<String>> processedMessages = this.preprocessAllLogs(logMessages);
+
+        this.calculateGroupTokenFreq(processedMessages);
+
+        Map<String, List<String>> logPatternMap = new HashMap<>();
+        for (List<String> processedMessage : processedMessages) {
+            String logId = processedMessage.getLast();
+            List<String> logPattern = this.parseLogPattern(processedMessage);
+            String patternKey = String.join(" ", logPattern);
+            logPatternMap.computeIfAbsent(patternKey, k -> new ArrayList<>()).add(logId);
```
Hailong-am, line 260: do we only need one record for each `patternKey`?

Hailong-am (approved):

Hailong-am:
build failure will handled by https://github.com/opensearch-project/skills/pull/489

PR #491, opened 2025-01-24
Add model related field for tools

xinyual:
### Description
To check downstream task before deleting, we need to add list of model related fields for model related tools. The background PR is: https://github.com/opensearch-project/ml-commons/pull/3209

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
+++ b/src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
@@ -6,6 +6,7 @@
 package org.opensearch.agent.tools;
 
 import static org.opensearch.action.support.clustermanager.ClusterManagerNodeRequest.DEFAULT_CLUSTER_MANAGER_NODE_TIMEOUT;
+import static org.opensearch.agent.tools.utils.CommonConstants.COMMON_MODEL_ID_FIELD;
 import static org.opensearch.ml.common.utils.StringUtils.isJson;
 
 import java.util.Arrays;
@@ -70,7 +71,6 @@ public class CreateAlertTool implements WithModelTool {
     @Getter
     private final String toolPrompt;
 
-    private static final String MODEL_ID = "model_id";
     private static final String PROMPT_FILE_PATH = "CreateAlertDefaultPrompt.json";
     private static final String DEFAULT_QUESTION = "Create an alert as your recommendation based on the context";
     private static final Map<String, String> promptDict = ToolHelper.loadDefaultPromptDictFromFile(CreateAlertTool.class, PROMPT_FILE_PATH);
@@ -301,7 +301,7 @@ public class CreateAlertTool implements WithModelTool {
 
         @Override
         public CreateAlertTool create(Map<String, Object> params) {
-            String modelId = (String) params.get(MODEL_ID);
+            String modelId = (String) params.get(COMMON_MODEL_ID_FIELD);
             if (org.apache.commons.lang3.StringUtils.isBlank(modelId)) {
                 throw new IllegalArgumentException("model_id cannot be null or blank.");
             }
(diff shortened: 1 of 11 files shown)


src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -66,10 +67,12 @@
 @Setter
 @Getter
 @ToolAnnotation(CreateAnomalyDetectorTool.TYPE)
-public class CreateAnomalyDetectorTool implements Tool {
+public class CreateAnomalyDetectorTool implements WithModelTool {
     // the type of this tool
     public static final String TYPE = "CreateAnomalyDetectorTool";
 
+    public static final String MODEL_ID_FIELD = "model_id";
```
zhichao-aws, line 74: Can we use this static variable to replace `"model_id"` constant in factory?

zhichao-aws (approved):

zhichao-aws:
It seems we have a compile error. Please fix it before merge.

```
/Users/runner/work/skills/skills/src/main/java/org/opensearch/agent/tools/VectorDBTool.java:115: error: Factory cannot be inherited with different arguments: <org.opensearch.agent.tools.NeuralSparseSearchTool> and <org.opensearch.agent.tools.VectorDBTool>
    public static class Factory extends AbstractRetrieverTool.Factory<VectorDBTool>
```

model-collapse (approved):

src/main/java/org/opensearch/agent/tools/CreateAlertTool.java
```diff
@@ -48,7 +48,7 @@
 
 @Log4j2
 @ToolAnnotation(CreateAlertTool.TYPE)
-public class CreateAlertTool implements Tool {
+public class CreateAlertTool implements WithModelTool {
```
Hailong-am, line 51: nit: instead of have a interface `WithModelTool`, i would recommend have another interface `ModelAwareResource`

```java

interface ModelAwareResource {
   String modelId();
}

public class CreateAlertTool implements Tool, ModelAwareResource {
```

Hailong-am (approved):

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
zane-neo: I see this constant been used in multiple classes, let's create a common constant class, eg. `CommonConstants` to put it there, and that class can be used to put more global constants in the future.

xinyual: > I see this constant been used in multiple classes, let's create a common constant class, eg. `CommonConstants` to put it there, and that class can be used to put more global constants in the future.

I think each tool should have its own related tool field so we cannot use a generate one for all tools

zane-neo: Not quite understand, I mean this constant: 
https://github.com/opensearch-project/skills/pull/491/files#diff-e7cc44d1e3c01541c76a3d3f5308487536871c349b84cf388baa9fb8c45a15a6R74 and 
https://github.com/opensearch-project/skills/pull/491/files#diff-cacec66258e0f1410446a7f762a38af26fb9d2874e562151820f0d0e55f95f89R36 and 
https://github.com/opensearch-project/skills/pull/491/files#diff-fa8cc057d2cd63132a2728fabab484d8b9b254faab603f50c189e50e12882d18R72
I see they're all string with value `model_id`, this can be placed to a common place. This is not a big change and if you think you can optimize this in future PRs, I'm fine to approve.

zane-neo (approved):

PR #676, opened 2025-12-02
new ad create enhanced tool to validate and create multiple detectors

amitgalitz:
### Description

This PR introduces CreateAnomalyDetectorToolEnhanced, a tool that leverages LLM capabilities to automatically generate, validate, and create anomaly detectors for different given indices

## Implementation Overview
The tool implements a step-by-step processing flow that handles batch operations on indices while maintaining robust error handling and validation mechanisms. 

### Code Flow
1. Extract & validate indices →
2. Get index mappings →
3. LLM generates detector config →
4. Validate detector structure →
5. Optimize parameters (suggest API) →
6. Validate model requirements →
7. Create & start detector

The implementation includes utilizing an LLM configuration generation, and retrying on different stages with the validate and suggest api to get optimal configuration that will lead to detector initializing
detector reliability.

The tool processes indices sequentially to prevent LLM throttling and implements retry logic with LLM-assisted error correction. Each index undergoes mapping analysis, LLM configuration generation, detector validation, suggest api based changes, and model validation before final deployment.

## Usage Example

Input:
```
{
  "indices": ["ecommerce-data", "server-logs"]
}
```

Output:
```
{
  "ecommerce-data": {
    "status": "success",
    "detectorId": "abc123",
    "detectorName": "ecommerce-data-detector-xyz",
    "createResponse": "Detector created successfully",
    "startResponse": "Detector started successfully"
  },
  "server-logs": {
    "status": "failed_validation",
    "error": "Insufficient data for model training"
  }
}
```

## Processing Flow
The implementation follows a step-by-step flow for each index: extraction and validation, mapping retrieval, LLM configuration generation, detector structure validation, parameter optimization via suggest API, model requirement validation, and finally detector creation and activation. The tool maintains state consistency across batch operations and writes it to the detector result object that is returned as a string so an LLM can use this as one of its tools

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List


src/main/java/org/opensearch/agent/tools/utils/AnomalyDetectorToolHelper.java
```diff
@@ -0,0 +1,68 @@
...
+            List<String> indices = (List<String>) input.get("indices");
+
+            if (indices == null || indices.isEmpty()) {
+                throw new IllegalArgumentException("No indices provided");
+            }
+
+            for (String index : indices) {
+                if (index.startsWith(".")) {
+                    throw new IllegalArgumentException("System indices not supported: " + index);
+                }
```
dbwiddis: I think we have [better detection of System indices](https://github.com/opensearch-project/OpenSearch/blob/f76826cc034f652a4cd99c51046a51f1ed6d0512/server/src/main/java/org/opensearch/indices/SystemIndices.java#L66-L82) than the leading dot.  [Example usage](https://github.com/opensearch-project/OpenSearch/blob/f76826cc034f652a4cd99c51046a51f1ed6d0512/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java#L448).

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+                return (String) content.getFirst().get("text");
+            } catch (Exception e) {
+                log.error("Failed to parse Claude content format", e);
+                return null;
+            }
+        } else {
+            log.error("Unknown response format. Available keys: {}", dataAsMap.keySet());
+            return null;
+        }
+    }
```
dbwiddis: This hard-coded parsing and relying on exception handling seems like it could be improved.  At least add a comment showing the expected format, so if this breaks in the future, the next maintainer (or future-you) can quickly identify the format change without walking through the code.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+        contextBuilder.append("Available fields:\n");
+        for (Map.Entry<String, String> field : mappingContext.filteredMapping.entrySet()) {
+            contextBuilder.append("- ").append(field.getKey()).append(": ").append(field.getValue()).append("\n");
+        }
+        contextBuilder.append("Available date fields: ").append(String.join(", ", mappingContext.dateFields)).append("\n\n");
+
+        String fixPrompt = createFixPrompt(originalSuggestions, validationError);
+        String fullPrompt = contextBuilder.toString() + fixPrompt;
+
+        callLLM(fullPrompt, tenantId, ActionListener.wrap(fixedResponse -> {
```
dbwiddis: Lots of duplicated code between `retryDetectorValidation()` and `retryModelValidation`, consider a helper method for the common parts.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+            ModelTensor modelTensor = modelTensors.getMlModelTensors().get(0);
+            Map<String, Object> dataAsMap = (Map<String, Object>) modelTensor.getDataAsMap();
+
+            if (dataAsMap == null) {
+                listener.onFailure(new IllegalStateException("Remote endpoint fails to inference."));
+                return;
+            }
+
+            String finalResponse = extractResponseFromDataAsMap(dataAsMap);
+            if (Strings.isNullOrEmpty(finalResponse)) {
```
dbwiddis: I like this, you should use it everwhere there's a `foo == null || foo.isEmpty()`

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+@Getter
+@ToolAnnotation(CreateAnomalyDetectorToolEnhanced.TYPE)
+public class CreateAnomalyDetectorToolEnhanced implements WithModelTool {
+    public static final String TYPE = "CreateAnomalyDetectorToolEnhanced";
+
+    private static final String DEFAULT_DESCRIPTION =
+        "Enhanced tool for creating anomaly detector configurations. Takes an index name, extracts the index mappings, and uses LLM to generate complete detector JSON configurations ready for the create detector API.";
+
+    private static final String EXTRACT_INFORMATION_REGEX =
+        "(?s).*\\{category_field=([^|]*)\\|aggregation_field=([^|]*)\\|aggregation_method=([^|]*)\\|interval=([^}]*)}.*";
```
dbwiddis: This seems to be a really strict regex that would fail on things like whitespace or the keys being in a different order (and cause retries which lose the benefit of the optimal date field.).  

Given we're just extracting characters between known strings, it seems there's likely a better way of iterating between the `|` characters, splitting on `=`, and trimming the key/value into a map?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+                            );
+                        } else {
+                            log.error("Max detector validation retries reached: {}", errorMessage);
+                            listener
+                                .onFailure(
+                                    new RuntimeException(
+                                        "Detector validation failed after " + MAX_DETECTOR_VALIDATION_RETRIES + " retries: " + errorMessage
+                                    )
+                                );
+                        }
```
dbwiddis: This conditional (`currentRetry >= MAX_DETECTOR_VALIDATION_RETRIES`) should be a fast-fail at the top of the `validateDetectorPhase` method, otherwise you spend a lot of time validating the detector which, if validation succeeds, you suddenly can't use.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+    ) {
+        log.info("Starting model validation");
+
+        callValidationAPI(detector, "model", new ActionListener<ValidateConfigResponse>() {
+            @Override
+            public void onResponse(ValidateConfigResponse response) {
+
+                if (response.getIssue() != null) {
+                    String issueAspect = response.getIssue().getAspect().toString();
+                    boolean isBlockingError = issueAspect != null && issueAspect.toLowerCase(Locale.ROOT).startsWith("detector");
```
dbwiddis: Is starting with "detector" always blocking?

Is there some helper method somewhere which defines blocking errors that you can use?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+                        retryModelValidation(detector, errorMessage, tenantId, maxRetries, currentRetry + 1, listener);
+                    } else {
+                        // Max retries reached
+                        if (isBlockingError) {
+                            log.error("Max retries reached with blocking validation error: {}", errorMessage);
+                            respondWithError(listener, detector.getIndices().get(0), "validation", errorMessage);
+                        } else {
+                            DetectorResult result = DetectorResult
+                                .failedValidation(detector.getIndices().get(0), "Non-blocking warning: " + errorMessage);
+                            listener.onResponse((T) result.toJson());
```
dbwiddis: If `detector.getIndices().get(0)` throws an exception the listener will never be completed. Probably need some exception handling here.

src/main/resources/org/opensearch/agent/tools/CreateAnomalyDetectorEnhancedPrompt.json
```diff
@@ -0,0 +1,4 @@
...
...\n* Real-time systems: 10-15 minutes (APIs, web services, errors, response times)\n* Infrastructure monitoring: 15-30 minutes (servers, databases, resource usage)\n* Business processes: 30-60 minutes (transactions, conversions)\n* Security logs: 10-30 minutes (access logs, firewalls, authentication)\n* Batch/ETL processes: 60+ minutes (data pipeline monitoring)\n* Sparse data: 60+ minutes (avoid false positives from empty buckets)\n\nSTEP 6 - SPARSE DATA INTELLIGENCE:\nCRITICAL: If validation suggests intervals >120 minutes due to sparse data:\n\n* PREFERRED SOLUTION: Remove category field entirely (better no segmentation than 4+ hour intervals)\n* ALTERNATIVE: Choose lower-cardinality category field\n* NEVER ACCEPT: If intervals are over 24 hours then we should aim for lower interval with different categorical fields.\n* PRINCIPLE: Operational usefulness over perfect segmentation\n\nOUTPUT FORMAT:\nReturn ONLY this structured format, no explanation:\n{category_field=field_name_or_empty|aggregation_field=field1,field2|aggregation_method=method1,method2|interval=minutes}\n\nIf no suitable fields exist, use empty strings for aggregation_field and aggregation_method.\n\nAssistant:\"",
```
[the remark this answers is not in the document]
kaituo: >e.g., user_agent.keyword, message.keyword

What if I want to monitor the message volume and the number of agents running?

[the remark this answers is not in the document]
kaituo: >category_field=field_name_or_empty

Should we tell LLM what happens when there is no categorical field?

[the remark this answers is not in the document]
kaituo: >interval=minutes

Should we tell LLM what happens when there are no suitable interval found?

src/main/resources/org/opensearch/agent/tools/CreateAnomalyDetectorEnhancedPrompt.json
```diff
@@ -0,0 +1,4 @@
...
...erational needs:\n\n* Real-time systems: 10-15 minutes (APIs, web services, errors, response times)\n* Infrastructure monitoring: 15-30 minutes (servers, databases, resource usage)\n* Business processes: 30-60 minutes (transactions, conversions)\n* Security logs: 10-30 minutes (access logs, firewalls, authentication)\n* Batch/ETL processes: 60+ minutes (data pipeline monitoring)\n* Sparse data: 60+ minutes (avoid false positives from empty buckets)\n\nSTEP 6 - SPARSE DATA INTELLIGENCE:\nCRITICAL: If validation suggests intervals >120 minutes due to sparse data:\n\n* PREFERRED SOLUTION: Remove category field entirely (better no segmentation than 4+ hour intervals)\n* ALTERNATIVE: Choose lower-cardinality category field\n* NEVER ACCEPT: If intervals are over 24 hours then we should aim for lower interval with different categorical fields.\n* PRINCIPLE: Operational usefulness over perfect segmentation\n\nOUTPUT FORMAT:\nReturn ONLY this structured format, no explanation:\n{category_field=field_name_or_empty|aggregation_field=field1,field2|aggregation_method=method1,method2|interval=minutes}\n\nIf no suitable fields exist, use empty strings for aggregation_field and aggregation_method"
```
kaituo: It is hard to maintain two prompts with almost identical content. Can you try to generate the claude one programmatically by adding extra information?

src/main/resources/org/opensearch/agent/tools/CreateAnomalyDetectorEnhancedPrompt.json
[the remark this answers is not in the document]
kaituo: >can immediately investigate and resolve

Can you define  what is the criteria by which the team can investigate and resolve?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorTool.java
```diff
@@ -335,21 +336,8 @@ private Map<String, String> enrichParameters(Map<String, String> parameters) {
...
-            String value = entry.getValue();
-            if (value.equals("date") || value.equals("date_nanos")) {
-                result.add(entry.getKey());
-            }
-        }
-        return result;
-    }
-
     @SuppressWarnings("unchecked")
+    **/
```
kaituo: @SuppressWarnings should be outside of comment block. This may caused CI compile failure.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+            listener.onFailure(e);
+        }));
+    }
+
+    private AnomalyDetector applySuggestionsToDetector(AnomalyDetector originalDetector, SuggestConfigParamResponse response) {
+        TimeConfiguration newInterval = originalDetector.getInterval();
+        TimeConfiguration newWindowDelay = originalDetector.getWindowDelay();
+        Integer newHistoryIntervals = originalDetector.getHistoryIntervals();
+        if (response.getInterval() != null) {
+            newInterval = response.getInterval();
```
kaituo: Would this override LLM's interval? If yes, is this on purpose? why do we need LLM's interval suggestion?

kaituo (commented):
partial review

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+            });
+    }
+
+    private String createFixPrompt(Map<String, String> originalSuggestions, String validationError) {
+        String currentInterval = originalSuggestions.getOrDefault("interval", "10");
+        String categoryField = originalSuggestions.get(OUTPUT_KEY_CATEGORY_FIELD);
+
+        // Check if this is a sparse data issue with unreasonably high suggested interval
+        boolean isUnreasonableInterval = validationError.contains("interval")
+            && (validationError.contains("240")
```
kaituo: Can we just say the interval is larger than 240, isUnreasonableInterval is true? Suggest API can give a larger number than 1440.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+                String bestDateField = futures
+                    .stream()
+                    .map(CompletableFuture::join)
+                    .max(Comparator.comparing(Map.Entry::getValue))
+                    .map(Map.Entry::getKey)
+                    .orElse(suggestedDateFields[0]); // fallback to first suggestion
+                listener.onResponse(bestDateField);
+            })
+            .exceptionally(e -> {
+                futures.forEach(FutureUtils::cancel);
```
kaituo: The futures may have already completed.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+                result = DetectorResult.failedValidation(indexName, errorMessage);
+                break;
+            case "create":
+                result = DetectorResult.failedCreate(indexName, errorMessage);
+                break;
+            case "start":
+                result = DetectorResult.failedStart(indexName, null, errorMessage);
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown status: " + status);
```
kaituo: Throw listener failure? Otherwise, this leaves the listener hanging.

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+     * 
+     * @param parameters Map containing "input" with JSON array of index names
+     * @param listener ActionListener to receive results as JSON string
+     */
+    @Override
+    public <T> void run(Map<String, String> parameters, ActionListener<T> listener) {
+        if (parameters.containsKey("input")) {
+            String inputStr = parameters.get("input");
+            if (inputStr != null && inputStr.trim().startsWith("[")) {
+                parameters.put("input", "{\"indices\": " + inputStr + "}");
```
kaituo: Should we sanitize user input? Is it possible we may have prompt injection from malicious users?

src/main/java/org/opensearch/agent/tools/CreateAnomalyDetectorToolEnhanced.java
```diff
@@ -0,0 +1,1402 @@
...
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            new IntervalTimeConfiguration(intervalMinutes, ChronoUnit.MINUTES),
```
kaituo: Frequency should be a random number between [2*interval, 24h] and a multiple of interval. Ideally, we can just put 24h as we don't require real time monitoring. But to have some jitter, we apply some random number. Otherwise, we risk breaking down the domain by running too many concurrent queries at the same time.

PR #752, opened 2026-06-22
Use PPL to fetch sample data in PPLTool

YiyuanMiao:
### Description
Use PPL to fetch sample data in `PPLTool` instead of a DSL `match_all` search.

Previously, `PPLTool` fetched a sample document via a DSL `match_all` search
(`client.search` with `MatchAllQueryBuilder`, size 1) to build the table info
that is fed into the LLM prompt. This change replaces that DSL search with a PPL
query (`source=<index> | head 1`) executed through `PPLExecuteHelper`, the shared
PPL execution utility already used by `LogPatternAnalysisTool`,
`DataDistributionTool`, and `DataFetchingHelper`.

This unifies sample retrieval with the existing PPL execution path (the final,
LLM-generated PPL is already run via PPL) and with the S3/Spark sample-handling
logic, and removes the OpenSearch-index-only DSL dependency for sample fetching.

Main changes:
- Replace the DSL `match_all` sample search with `PPLExecuteHelper.executePPLAndParseResult`
  using `source=<index> | head 1`.
- Add `extractSampleFromPPLResult` to recombine the PPL jdbc `schema` (column
  names) and `datarows` (values) into a field-name → value map, reusing the
  existing sample-extraction logic.
- Change `constructTableInfo` to accept the parsed sample map instead of `SearchHit[]`.
- Remove the now-unused `buildSearchRequest` method and the DSL-only imports
  (`SearchRequest`, `SearchSourceBuilder`, `MatchAllQueryBuilder`, `SearchHit`).

Tests:
- Update the existing `PPLToolTests` to mock the sample-fetch PPL query
  (`source=demo | head 1`) and differentiate it from the final PPL execution by
  query string, returning a jdbc-formatted result for the sample query.
- Add `testTool_sampleValuesFromPPLReachPrompt` to verify the sample values
  fetched via PPL actually flow into the prompt sent to the LLM.
- Adjust the error-path tests to assert on the exception **message** (matching the
  pattern used in `LogPatternAnalysisToolTests`) rather than the exact exception
  type, since errors now propagate through `PPLExecuteHelper`.

### Why the existing tests were updated

  The change replaces sample-data fetching in `PPLTool` from a DSL `match_all`
  search (`client.search`) to a PPL query run through `PPLExecuteHelper`
  (`client.execute(PPLQueryAction)`). Because every affected test exercises the
  sample-fetch step, both their mocking and their assertions had to be adjusted.
  The updates fall into two categories.

  #### Category 1: Mocking the sample-fetch path (happy-path tests)

  - **Before:** sample data was fetched via `client.search`, so the tests mocked
    `client.search` to return a `SearchResponse` / `SearchHit[]`.
  - **After:** sample data is fetched via `client.execute(PPLQueryAction)`, so the
    tests now mock the PPL execution instead.
  - Since the sample-fetch query and the final (LLM-generated) PPL execution now go
    through the **same** `PPLQueryAction`, the mock differentiates them by the PPL
    query string: the sample query (`source=demo | head 1`) returns a jdbc-formatted
    result, while the final execution returns the plain result.

  #### Category 2: Adjusting exception assertions (error-path tests)

  Affected tests: `testToolWhenGettingSagemakerError`,
  `testTool_withWrongEndpointInference`,
  `testTool_withWrongEndpointInferenceWithNullResponse`, `testTool_executePPLFailure`.

  - **Before:** these tests asserted on the exact exception **type**
    (`OpenSearchStatusException`, `IllegalStateException`).
  - **After:** errors now propagate through `PPLExecuteHelper`, which wraps them in a
    `RuntimeException`, so asserting the exact type no longer holds.
  - The tests were changed to assert on the exception **message** instead (e.g.
    `<SAGEMAKER_ENDPOINT>`, `Remote endpoint fails to inference.`), matching the
    pattern already used in `LogPatternAnalysisToolTests` for errors surfaced through
    `PPLExecuteHelper`.

  **No functional regression:** the SageMaker redaction logic still works — the
  redacted placeholder `<SAGEMAKER_ENDPOINT>` is present and the raw ARN / account ID
  are removed. The only change is that the exception type is now wrapped by
  `PPLExecuteHelper`, consistent with how other tools using the same helper behave.

### Updated error-path tests

  All four tests are in the same file: `src/test/java/org/opensearch/agent/tools/PPLToolTests.java`

  | Test | File | Purpose | Simulated failure | How it was changed |
  |----|----|-------------|-------------------|--------------------|
  | `testToolWhenGettingSagemakerError` | `PPLToolTests.java` | Verify that when SageMaker returns an error, sensitive info (ARN, account ID) is redacted | LLM call fails with an `OpenSearchStatusException` containing a SageMaker ARN | Assert the exception **message** contains `<SAGEMAKER_ENDPOINT>` and no longer contains the raw ARN / account ID, instead of asserting the exact exception type |
  | `testTool_withWrongEndpointInference` | `PPLToolTests.java` | Verify the tool reports an error when the LLM returns an error code instead of a PPL | `{"code":"424"}` (no `response` field) | Assert the exception message contains `"Remote endpoint fails to inference."` instead of asserting the exact exception type |
  | `testTool_withWrongEndpointInferenceWithNullResponse` | `PPLToolTests.java` | Verify the tool reports an error when the LLM returns a null response | `{"response": null}` | Same as above — assert the message contains `"Remote endpoint fails to inference."` |
  | `testTool_executePPLFailure` | `PPLToolTests.java` | Verify the tool reports an error when executing the final (LLM-generated) PPL fails | The final PPL execution throws an exception | Differentiate the two PPL calls via `isSamplePPL`: let the sample-fetch query succeed and only the final execution fail; assert the message contains `"execute ppl error"` |

### Check List
- [ ] API changes companion pull request created.
- [ ] Public documentation issue/PR created.

### DIFF
--- a/src/main/java/org/opensearch/agent/tools/PPLTool.java
+++ b/src/main/java/org/opensearch/agent/tools/PPLTool.java
@@ -42,13 +42,12 @@ import org.opensearch.OpenSearchStatusException;
 import org.opensearch.Version;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.admin.indices.mapping.get.GetMappingsRequest;
-import org.opensearch.action.search.SearchRequest;
+import org.opensearch.agent.tools.utils.PPLExecuteHelper;
 import org.opensearch.agent.tools.utils.ToolHelper;
 import org.opensearch.agent.tools.utils.mergeMetaData.MergeRuleHelper;
 import org.opensearch.cluster.metadata.MappingMetadata;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.action.ActionResponse;
-import org.opensearch.index.query.MatchAllQueryBuilder;
 import org.opensearch.ml.common.FunctionName;
 import org.opensearch.ml.common.dataset.remote.RemoteInferenceInputDataSet;
 import org.opensearch.ml.common.input.MLInput;
@@ -60,8 +59,6 @@ import org.opensearch.ml.common.spi.tools.WithModelTool;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskAction;
 import org.opensearch.ml.common.transport.prediction.MLPredictionTaskRequest;
 import org.opensearch.ml.common.utils.ToolUtils;
-import org.opensearch.search.SearchHit;
-import org.opensearch.search.builder.SearchSourceBuilder;
 import org.opensearch.sql.plugin.transport.PPLQueryAction;
 import org.opensearch.sql.plugin.transport.TransportPPLQueryRequest;
 import org.opensearch.sql.plugin.transport.TransportPPLQueryResponse;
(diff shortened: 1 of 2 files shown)


Hailong-am:
depends on https://github.com/opensearch-project/skills/pull/758 to fix compile error

src/main/java/org/opensearch/agent/tools/PPLTool.java
```diff
@@ -565,17 +568,44 @@ private String constructTableInfoByPPLResultForSpark(Map<String, Object> schema,
...
+        if (!(schemaObj instanceof List) || !(datarowsObj instanceof List)) {
+            return sampleSource;
+        }
+        List<Map<String, Object>> schema = (List<Map<String, Object>>) schemaObj;
+        List<List<Object>> datarows = (List<List<Object>>) datarowsObj;
+        if (datarows.isEmpty()) {
+            return sampleSource;
+        }
+        List<Object> firstRow = datarows.get(0);
+        for (int i = 0; i < schema.size() && i < firstRow.size(); i++) {
```
Hailong-am, line 588: nit: is there a case `schema.size()` different with `firstRow`

Hailong-am (approved):

PR #783, opened 2026-08-17
Use Eclipse P2 mirror (skills)

peterzhuamazon:
Routes Spotless Eclipse P2 downloads through https://ci.opensearch.org/ to avoid direct download.eclipse.org outages.

Tracking bulletin: https://github.com/opensearch-project/opensearch-build/issues/6421#issuecomment-5271186726


peterzhuamazon:
expected mirror update.

